PR #8, opened 2016-12-23
Updates for release of BouncyCastle 1.56

peterdettman:
- add maven entry for BC 1.56 and update the versions range
- DHIES weak crypto tests now pass if algs not even supported
- some ECIES entries moved to the invalid list
- ECIES decryption init calls fixed to pass the correct parameters
- EciesTest.testModifyPoint now expects GeneralSecurityException

### DIFF
--- a/java/com/google/security/wycheproof/testcases/EciesTest.java
+++ b/java/com/google/security/wycheproof/testcases/EciesTest.java
@@ -17,6 +17,7 @@
 package com.google.security.wycheproof;
 
 import java.nio.ByteBuffer;
+import java.security.GeneralSecurityException;
 import java.security.KeyPair;
 import java.security.KeyPairGenerator;
 import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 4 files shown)


thaidn:
Cool, thanks for working on this!

I'll merge this PR after setting up two-way sync between GitHub and our internal repository, stay tuned.

java/com/google/security/wycheproof/testcases/EciesTest.java
```diff
@@ -296,7 +299,7 @@ public void testIsAlias(String algorithmA, String algorithmB) throws Exception {
     byte[] message = "Hello".getBytes("UTF-8");
     eciesA.init(Cipher.ENCRYPT_MODE, keyPair.getPublic());
     byte[] ciphertext = eciesA.doFinal(message);
-    eciesB.init(Cipher.DECRYPT_MODE, keyPair.getPrivate());
+    eciesB.init(Cipher.DECRYPT_MODE, keyPair.getPrivate(), eciesA.getParameters());
```
thaidn: Should this be eciesB.getParameters()?

peterdettman: @thaidn Sorry for the delayed reply. Actually eciesA.getParameters() was correct, since we are trying to decrypt with the parameters that were used (possibly auto-generated) during encryption. Looking at current code it appears this was actually changed to eciesB, which ought to be reverted (there's currently a test failure with latest BC caused by using eciesB here).

PR #9, opened 2016-12-27
Add testMaxPlaintextLength to check GCM plaintext limits

SalusaSecondus:
This adds a test to ensure implementations enforce maximum plaintext lengths as defined by NIST SP 800-38D.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -32,7 +32,8 @@ load(":build_defs.bzl", "bouncycastle_all_tests", "spongycastle_all_tests")
 # $ bazel test BouncyCastleAllTest_*
 bouncycastle_all_tests(
     # This test takes a long time, because key generation for DSA and DH generate new parameters.
-    size = "large",
+    # Also, it encrypts several gigabytes of test data.
+    size = "enormous",
     srcs = ["java/com/google/security/wycheproof/BouncyCastleAllTests.java"] + test_srcs,
     test_class = "com.google.security.wycheproof.BouncyCastleAllTests",
     deps = common_deps,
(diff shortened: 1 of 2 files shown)


SalusaSecondus:
I signed it! (I'm covered by Amazon.com's corporate SLA.)

quannguyentm:
Thanks SalusaSecondus, this is one of few @SlowTest tests that we haven't published yet and we'll publish them after the holidays. 
FYI, I found this GCM bug in Bouncy Castle (CVE-2015-6644) and OpenJDK8 which lead to the following fixes: 
BouncyCastle (https://android.googlesource.com/platform/external/bouncycastle/+/3e128c5fea3a0ca2d372aa09c4fd4bb0eadfbd3f%5E%21/#F0)
OpenJDK8 (http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/rev/0c3ed12cdaf5)

SalusaSecondus:
So you're the person to thank. Thank you! I did check to ensure the libraries were patched before I made this test public.

I'm starting to think we may need a better way to coordinate to avoid duplicate work. I've a few more tests that I want to draft for AesGcm (and possibly AesEax) but don't know what will pass yet and don't want to write tests you've already spent time on.

thaidn:
The test that Quan alluded to earlier has been released: https://github.com/google/wycheproof/blob/master/java/com/google/security/wycheproof/testcases/AesGcmTest.java#L452.

Please reopen this PR if you think that test is not quite the same as yours.

Thanks!

Issue #4, opened 2016-12-20
Please make it possible to run tests on local jars


mouse07410:
Current wycherproof tries to download BouncyCastle from the Internet. I need wycherproof to run tests on the local BouncyCastle jars.

How can I do it?

thaidn:
Emilia, could you please take a look? Not urgent, but when you have cycles.

mouse07410:
I think that far more people would like/need to test the stuff they have locally, compared to those who want to test what's in the global repo.

Thanks!

peterdettman:
In the absence of an actual test target, you can at least edit the WORKSPACE file to configure a jar as a local file:

```
http_jar(
    name = "bouncycastle_1_56",
    url = "file:///Users/peter/Downloads/bcprov-jdk15on-156.jar"
)
```

You could just use an existing target and point its jar locally as above, or edit build_defs.bzl to add a separate target referencing the jar.

Buantai:
Aw
aresome

mouse07410:
I must be dense...

I've edited `WORKPSPACE` file (just removing the 1.48-1.54 related stuff at the top):
```
#maven_jar(
#    name = "bouncycastle_1_55",
#    artifact = "org.bouncycastle:bcprov-jdk15on:1.55",
#)

http_jar(
    name = "bouncycastle_1_56",
    url = "file:///Library/Java/Extensions/bcprov-ext-jdk15on-156.jar"
)
. . . . .
```

My attempt to do `bazel test BouncyCastleAllTests` produced:
```
$ bazel test BouncyCastleAllTests
Warning: ignoring http_proxy in environment.
..............
ERROR: /Users/ur20980/Src/wycheproof/BUILD:33:1: no such package '@bouncycastle_1_55//jar': error loading package 'external': The repository named 'bouncycastle_1_55' could not be resolved and referenced by '//:BouncyCastleAllTests'.
ERROR: Analysis of target '//:BouncyCastleAllTests' failed; build aborted.
INFO: Elapsed time: 4.864s
ERROR: Couldn't start the build. Unable to run tests.
$ 
```

Frankly, I am not aspiring to become a wycherproof wizard - being able to (smoothly) run it as a black box tool over my BouncyCastle jars (and later - over C++ crypto libraries like Crypto++) is all I hope for.

mouse07410:
**Update2 - got something to run**

1. Did not remove the already-defined BouncyCastle targets - *added* `http_jar` for 1.56.
[...]

PR #10, opened 2016-12-28
Option to test a local BouncyCastle jar

ekasper:
Resolves Github issue #4

This is a bit of a hack. I can't find a way to supply the jar directly
from the command line to the 'deps' or 'srcs' of a rule, so we create
an intermediate symlink.

Also, we need to define a separate build target. It'd be nicer to use
config_setting to switch between local and remote jars - but macros
don't support command-line flags, which makes this awkward.


ekasper:
Closing in favour of https://github.com/google/wycheproof/pull/16

PR #20, opened 2017-02-01
Set tags=[manual] for *Local targets, thus exclude them from the wildcard target patterns

thaidn:


thaidn:
This should fix #4.

BUILD
```diff
@@ -40,6 +40,9 @@ bouncycastle_all_tests(
 
 java_test(
     name = "BouncyCastleAllTestsLocal",
+    # this target requires specifing a shell variable, thus won't work with the wildcard target patterns.
```
ekasper: typo: specifying (x2)

PR #29, opened 2017-06-01
Add tests for partially-overlapping input/output buffers

bdonlan:
update/doFinal are required to be "copy-safe". From the javadoc:

    Note: this method should be copy-safe, which means the input and output
    buffers can reference the same block of memory and no unprocessed input
    data is overwritten when the result is copied into the output buffer.

However many implementations of JCE methods do not implement this property in
all cases. Notably, BouncyCastle fails this test on arrays, and the built-in
OpenJDK providers fail when the input bytebuffer is either a read-only
array-backed byte buffer, or if the buffers are direct buffers. For SunJCE,
this has been acknowledged publicly by Oracle at
https://bugs.openjdk.java.net/browse/JDK-8181386 .

These new tests verify correct overlapping buffer handling for AES/GCM
specifically; as future work, it would be good to generalize these tests for
other ciphers and modes.

### DIFF
--- a/java/com/google/security/wycheproof/testcases/AesGcmTest.java
+++ b/java/com/google/security/wycheproof/testcases/AesGcmTest.java
@@ -16,6 +16,8 @@
 
 package com.google.security.wycheproof;
 
+import static org.junit.Assert.assertArrayEquals;
+
 import com.google.security.wycheproof.WycheproofRunner.ExcludedTest;
 import com.google.security.wycheproof.WycheproofRunner.ProviderType;
 import com.google.security.wycheproof.WycheproofRunner.SlowTest;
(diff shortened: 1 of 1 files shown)


bdonlan:
Note that both BC and SunJCE (on both 8u131 and 9-ea) fail the new testByteBufferShiftedAlias test.

java/com/google/security/wycheproof/testcases/AesGcmTest.java
```diff
@@ -268,6 +270,167 @@ public void testByteBufferAlias() throws Exception {
...
+          assertArrayEquals(ptVector,
+                            Arrays.copyOfRange(outBuf, outputOffsetInBuffer, outputOffsetInBuffer + resultPtLength));
+        } catch (Throwable t) {
+          throw new AssertionError("testLargeByteBufferAlias failed with outputOffset=" + outputOffset, t);
+        }
+      }
+    }
+  }
+
+  /**
```
quannguyentm: Thanks for your pull request. Can you please add https://bugs.openjdk.java.net/browse/JDK-8181386 which has more detailed description of the bug to the comment?

java/com/google/security/wycheproof/testcases/AesGcmTest.java
```diff
@@ -268,6 +270,169 @@ public void testByteBufferAlias() throws Exception {
...
+
+            inputRO = input.duplicate();
+            break;
+          }
+          default: {
+            throw new AssertionError("Unknown test index " + i);
+          }
+        }
+
+        // Now early points to all but the last byte of the buffer, and late and lateRead both point to all but the
```
bdonlan: Oops, I think that's left over from before I refactored this code. I'll rewrite these comments.

java/com/google/security/wycheproof/testcases/AesGcmTest.java
```diff
@@ -268,6 +270,169 @@ public void testByteBufferAlias() throws Exception {
...
+
+        byte[] inBuf = new byte[sliceLength + Math.max(inputOffsetInBuffer, outputOffsetInBuffer)];
+        byte[] outBuf = inBuf;
+
+        System.arraycopy(ptVector, 0, inBuf, inputOffsetInBuffer, ptVector.length);
+
+        try {
+          int ctLength = 0;
+          if (useUpdate > 0) {
+            ctLength += cipher.update(inBuf, inputOffsetInBuffer, ptVector.length, outBuf, outputOffsetInBuffer);
```
bdonlan: Fixing.

java/com/google/security/wycheproof/testcases/AesGcmTest.java
```diff
@@ -268,6 +270,169 @@ public void testByteBufferAlias() throws Exception {
...
+
+          input.put(tmp);
+          input.flip();
+          inputRO.limit(input.limit());
+
+          cipher.init(Cipher.DECRYPT_MODE, key, parameters);
+          cipher.doFinal(inputRO, output);
+
+          output.flip();
+          assertEquals(ByteBuffer.wrap(ptVector), output);
```
bdonlan: 1. ByteBuffer equality is defined in terms of byte contents, and is not sensitive to the specific values of capacity/etc, except inasmuch that two bytebuffers with different `remaining()` values will be nonequal. See [the javadoc](https://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html#equals(java.lang.Object\)):

> Tells whether or not this buffer is equal to another object.
> 
> Two byte buffers are equal if, and only if,
> 
>     1. They have the same element type,
> 
>     2. They have the same number of remaining elements, and
> 
>     3. The two sequences of remaining elements, considered independently of their starting positions, are pointwise equal.
> 
> A byte buffer is not equal to any other type of object. 

2. Your assert succeeds because `byteBufferToHex()` performs an implicit `flip()` - which, since position() == 0 for both inputs, results in comparing two empty buffers.

PR #39, opened 2017-08-25
Fixed #38 - JUnit3 runner is being used instead of JUnit4

### LINKED ISSUE #38: JUnit3 runner is being used instead of JUnit4
The JUnit3 runner is being used instead of JUnit4.
This causes the annotations (Slow/Fast) to be ignored.
I have a fix and will submit a pull request shortly.

There was a [timeout issue](https://groups.google.com/forum/#!topic/wycheproof-users/usPJedwMPkw) reported on the mailing list in Feb.  It was identified then that SlowTest annotations were being ignored. I believe this was likely the cause of that issue.

**Repro Steps:**
- In junit-4.12.jar!/org/junit/internal/builders/AllDefaultPossibilitiesBuilder.class, set a breakpoint in the `return runner` line of the runnerForClass method
- Set a breakpoint inside of testWrappedAroundCounter (a SlowTest)
- Run a test with debugger attached
i.e. bazel test --java_debug  --genrule_strategy=standalone --spawn_strategy=standalone BouncyCastleTest_1_57

**Actual Result:**
- The breakpoint will show JUnit38ClassRunner being initialized
- The breakpoint in testWrappedAroundCounter will be hit

**Expected Result:**
- A JUnit4 runner is initialized
- testWrappedAroundCounter does **not** run during a 'Fast' test

**Fix:**
- Remove `extends TestCase` which triggers usage of the JUnit3 runner
- Replace `import static org.junit.Assert.assertArrayEquals;` with `import static org.junit.Assert.*;`
- Replace `import junit.framework.TestCase;` with `import org.junit.Test;`
- Annotate all existing test methods (that are prefixed with 'test' and have no args) with `@Test`

**Verify Fix:**
- Use same breakpoints as before in AllDefaultPossibilitiesBuilder and AesGcmTest.
- Run the same 'Fast' test
- The breakpoint in AllDefaultPossibilitiesBuilder.runnerForClass will show runner as BlockJunit4ClassRunner (instead of JUnit38ClassRunner)
- The breakpoint in testWrappedAroundCounter will not be hit.

jordanmews:
- Remove `extends TestCase` which triggered usage of the JUnit3 runner
- Replace `import static org.junit.Assert.assertArrayEquals;` with `import static org.junit.Assert.*;`
- Replace `import junit.framework.TestCase;` with `import org.junit.Test;`
- Annotate all existing test methods (that are prefixed with 'test' and have no args) with `@Test`

### DIFF
--- a/java/com/google/security/wycheproof/testcases/AesEaxTest.java
+++ b/java/com/google/security/wycheproof/testcases/AesEaxTest.java
@@ -22,13 +22,15 @@
 
 package com.google.security.wycheproof;
 
+import static org.junit.Assert.*;
+
 import javax.crypto.Cipher;
 import javax.crypto.spec.GCMParameterSpec;
 import javax.crypto.spec.SecretKeySpec;
-import junit.framework.TestCase;
+import org.junit.Test;
 
 /** AES-EAX tests */
-public class AesEaxTest extends TestCase {
+public class AesEaxTest {
 
   /** Test vectors */
   public static class EaxTestVector {
@@ -264,6 +266,7 @@ public class AesEaxTest extends TestCase {
     }
   }
 
+  @Test
   public void testLateUpdateAAD() throws Exception {
     for (EaxTestVector test : EAX_TEST_VECTOR) {
       Cipher cipher = Cipher.getInstance("AES/EAX/NoPadding");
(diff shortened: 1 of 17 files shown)


quannguyentm:
Thanks for your fix :)

PR #102, opened 2024-01-06
Fix computed tag in Ascon-80pq

### LINKED ISSUE #101: Tag in Ascon-80pq test vector is incorrect
Hi,

I wanted to use the Wycheproof to test against my implementation of the Ascon cipher. I was already using the KAT test vectors by [pyascon](https://github.com/meichlseder/pyascon/blob/master/genkat.py) (one of the reference implementations) so I was caught by surprise that the computed tag for Ascon-80pq didn't match with what Wycheproof was expecting. All of the [Pseudorandom flagged test vectors](https://github.com/google/wycheproof/blob/d9f6ec7d8bd8c96da05368999094e4a75ba5cb3d/testvectors_v1/ascon80pq_test.json#L31-L1206) were failing due to the mismatch of the tag, the ciphertext on the other hand was correct. I thought this was a fault in my implementation, but testing these vectors against the reference implementations of Ascon, didn't yield the result that Wycheproof expects (for example, [test vector 84](https://github.com/google/wycheproof/blob/d9f6ec7d8bd8c96da05368999094e4a75ba5cb3d/testvectors_v1/ascon80pq_test.json#L1193-L1206)):

<details>
<summary>C implementation</summary>

Using https://github.com/ascon/ascon-c
```c
#include <stdio.h>
#include <string.h>

#include "api.h"
#include "crypto_aead.h"

void print(unsigned char c, unsigned char *x, unsigned long long xlen) {
  unsigned long long i;
  printf("%c[%d]=", c, (int)xlen);
  for (i = 0; i < xlen; ++i) printf("%02x", x[i]);
  printf("\n");
}

int main() {
  unsigned char n[CRYPTO_NPUBBYTES] = {36,  100, 126, 227, 13, 38, 50, 237,
                                       203, 225, 95,  60,  28, 26, 32, 71};
  unsigned char k[CRYPTO_KEYBYTES] = {84,  1,   235, 204, 159, 186, 226,
                                      210, 50,  126, 126, 98,  129, 53,
                                      182, 179, 243, 6,   161, 132};
  unsigned char a[65] = {21,  151, 43,  4,   224, 64,  173, 141, 125, 219, 147,
                         237, 43,  214, 130, 92,  54,  50,  187, 53,  203, 50,
                         219, 178, 103, 203, 92,  174, 169, 206, 64,  98,  155,
                         214, 229, 59,  191, 247, 171, 93,  36,  173, 107, 213,
                         175, 63,  182, 193, 134, 98,  65,  255, 17,  255, 32,
                         140, 59,  185, 202, 175, 52,  249, 189, 78,  235};
  unsigned char m[65] = {21,  87,  96,  159, 102, 213, 162, 235, 120, 59,  131,
                         246, 22,  124, 223, 62,  186, 233, 46,  122, 133, 3,
                         89,  60,  203, 239, 128, 33,  72,  110, 207, 186, 75,
                         255, 20,  89,  15,  226, 205, 184, 57,  217, 48,  38,
                         130, 106, 111, 189, 130, 250, 181, 166, 175, 28,  108,
                         181, 75,  194, 14,  33,  131, 136, 238, 213, 129};
  unsigned char c[65 + 32], t[32];
  unsigned long long alen = 65;
  unsigned long long mlen = 65;
  unsigned long long clen = CRYPTO_ABYTES;
  int result = 0;
  
    print('k', k, CRYPTO_KEYBYTES);
  printf(" ");
  print('n', n, CRYPTO_NPUBBYTES);
  printf("\n");
  print('a', a, alen);
  printf(" ");
  print('m', m, mlen);
  printf(" -> ");
  result = crypto_aead_encrypt(c, &clen, m, mlen, a, alen, (void *)0, n, k);
  print('c', c, clen - CRYPTO_ABYTES);
  printf(" ");
  print('t', c + clen - CRYPTO_ABYTES, CRYPTO_ABYTES);
  return result;
  ```
</details>

<details>
<summary>Python implementation</summary>

Using https://github.com/meichlseder/pyascon
```python
import ascon

key = bytes.fromhex('5401ebcc9fbae2d2327e7e628135b6b3f306a184')
nonce = bytes.fromhex('24647ee30d2632edcbe15f3c1c1a2047')
associateddata = bytes.fromhex('15972b04e040ad8d7ddb93ed2bd6825c3632bb35cb32dbb267cb5caea9ce40629bd6e53bbff7ab5d24ad6bd5af3fb6c1866241ff11ff208c3bb9caaf34f9bd4eeb')
plaintext = bytes.fromhex('1557609f66d5a2eb783b83f6167cdf3ebae92e7a8503593ccbef8021486ecfba4bff14590fe2cdb839d93026826a6fbd82fab5a6af1c6cb54bc20e218388eed581')

ciphertext = ascon.encrypt(key, nonce, associateddata, plaintext, "Ascon-80pq")
    
print(f"key: {key.hex()}")
print(f"plaintext: {plaintext.hex()}")
print(f"ass.data: {associateddata.hex()}")
print(f"ciphertext: {ciphertext[:-16].hex()}")
print(f"tag: {ciphertext[-16:].hex()}")
```
</details>

Both implementations (and my implementation) say that `ccd345de03169a3e5c2cc27c58c43a62` is the correct tag instead of what Wycheproof claims, `e32cd6424cab0c59c528db6f70b81a86`. Are the reference implementations wrong here, or is the Wycheproof wrong here?

Gusted:
Resolves #101.

### DIFF
--- a/testvectors_v1/ascon80pq_test.json
+++ b/testvectors_v1/ascon80pq_test.json
@@ -39,7 +39,7 @@
           "aad" : "",
           "msg" : "",
           "ct" : "",
-          "tag" : "5b5706c1d4ba053e9d1173a7d28d8170",
+          "tag" : "a781b81a25c11b25197db620d344dac2",
           "result" : "valid"
         },
         {
@@ -53,7 +53,7 @@
           "aad" : "",
           "msg" : "d9",
           "ct" : "c1",
-          "tag" : "73764302c543cb9b8311ea801a00a336",
+          "tag" : "60b6692f2f80c1ca8c0f28fe99a60ece",
           "result" : "valid"
         },
         {
@@ -67,7 +67,7 @@
           "aad" : "",
           "msg" : "e0d1",
           "ct" : "6f6b",
-          "tag" : "841b97b9b418a3faddd9782ea5c298dc",
+          "tag" : "9b8770eaa636ad501f332b841a35e71f",
           "result" : "valid"
         },
         {
@@ -81,7 +81,7 @@
           "aad" : "",
           "msg" : "14c495",
           "ct" : "cd3414",
-          "tag" : "e99bf441c995b6a0f48dd4642c313cee",
+          "tag" : "413b422450550cd7341d1e609a78329b",
           "result" : "valid"
         },
         {
@@ -95,7 +95,7 @@
           "aad" : "",
           "msg" : "3e5f3f5c",
           "ct" : "60712f24",
-          "tag" : "29cc8355ee75bd192988aae1bbf01b58",
+          "tag" : "fa0bbda5fec463fe52aa1c00a6a6997f",
           "result" : "valid"
         },
         {
@@ -109,7 +109,7 @@
           "aad" : "",
           "msg" : "150cb5509f",
           "ct" : "d8418513fd",
-          "tag" : "b995af11dc566b26b5c48a73042aefef",
+          "tag" : "359dbb3f1b251cea21633047084a4941",
           "result" : "valid"
         },
         {
@@ -123,7 +123,7 @@
           "aad" : "",
           "msg" : "e582b374f6b5",
           "ct" : "e626d009a578",
-          "tag" : "fa3265f56658f80f4896b5f0c38fe33f",
+          "tag" : "3341d19b1e60dd2564c5066b46804581",
           "result" : "valid"
         },
         {
@@ -137,7 +137,7 @@
           "aad" : "",
           "msg" : "2a4e98eb59ce93",
           "ct" : "59401206f428cf",
-          "tag" : "36a4604c4f92c00ea7a339f90d6930be",
+          "tag" : "24741812d8e6d3acc43a6a0d896b0297",
           "result" : "valid"
         },
         {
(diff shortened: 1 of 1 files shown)


Gusted:
> finalize in Ascon80pq uses the last 16 bytes of the key, which is something that might get overlooked

Just to clarify as I was kind of confused by this comment, you mean that the last 16 bytes of the key is xor'red against the last 16 bytes of the state to compute the tag. As the whole key is also xorr'ed against the state before the permutation in the finalization step.

> But of course, this has to be fixed in the test vector generation code.

Is this code public?

bleichenbacher-daniel:
> > finalize in Ascon80pq uses the last 16 bytes of the key, which is something that might get overlooked
> 
> Just to clarify as I was kind of confused by this comment, you mean that the last 16 bytes of the key is xor'red against the last 16 bytes of the state to compute the tag. As the whole key is also xorr'ed against the state before the permutation in the finalization step.

Yes, indeed. E.g., Ascon v.1.1 does not specify that the last 16 bytes are being used to compute the tag, because this version does not include Ascon80pq and hence does not have to specify which bytes to use. Hence taking an old implementation and carelessly extending it to Ascon80pq can lead to such problems.
> 
> > But of course, this has to be fixed in the test vector generation code.
> 
> Is this code public?

No, it is not. Having access to it would solve this issue much faster, of course. For example it probably would document the version of the paper that was used for the implementation.

chuckx:
> Thanks a lot for doing this. I have a new implementation of Ascon, which confirms the tags above.

Given this confirmation, I'll go ahead and get this merged.

PR #109, opened 2024-03-28
Package wycheproof tests for the PHP ecosystem

tob-scott-a:
This does the absolute minimum to publish Wycheproof to the PHP ecosystem.

After merging this PR, the C2SP maintainers will need to publish this repository to Packagist to make it usable: https://packagist.org/about

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,22 @@
+{
+    "name": "c2sp/wycheproof",
+    "description": "Wycheproof test vectors for cryptographic software",
+    "authors": [
+        {
+            "name": "Scott Arciszewski",
+            "email": "scott.arciszewski@trailofbits.com",
+            "homepage": "https://github.con/tob-scott-a"
+        }
+    ],
+    "type": "library",
+    "keywords": ["wycheproof", "testing", "cryptography", "crypto"],
+    "license": "Apache-2.0",
+    "autoload": {
+        "psr-4": {
+            "C2sp\\Wycheproof\\": "php"
+        }
+    },
+    "require": {
+        "ext-json": "*"
+    }
+}
--- a/php/README.md
+++ b/php/README.md
@@ -1,0 +1,9 @@
+# Wycheproof for PHP
+
+This does not implement a complete testing harness for PHP software.
+This code exists to expose Wycheproof test vectors to the PHP ecosystem and 
+provide a simple way to get the file paths for all test vectors.
+
+Users are responsible for selecting the test vector file they are interested in,
+decoding the JSON file, iterating over the objects, and interacting with their
+chosen cryptography implementation.
--- a/php/Wycheproof.php
+++ b/php/Wycheproof.php
@@ -1,0 +1,48 @@
+<?php
+declare(strict_types=1);
+namespace C2sp\Wycheproof;
+
+class Wycheproof
+{
+    const DEFAULT_TEST_VECTOR_DIR_NAME = 'testvectors_v1';
+
+    private string $testVectorRoot;
+
+    public function __construct(?string $testVectorRoot = null)
+    {
+        if (is_null($testVectorRoot)) {
+            // Default to where ever the test vectors are installed locally
+            $testVectorRoot = dirname(__DIR__) .
+                DIRECTORY_SEPARATOR .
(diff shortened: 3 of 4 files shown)


composer.json
```diff
@@ -0,0 +1,22 @@
+{
+    "name": "c2sp/wycheproof",
+    "description": "Wycheproof test vectors for cryptographic software",
+    "authors": [
+        {
+            "name": "Scott Arciszewski",
+            "email": "scott.arciszewski@trailofbits.com",
+            "homepage": "https://github.con/tob-scott-a"
```
reaperhulk, line 8:
```suggestion
            "homepage": "https://github.com/tob-scott-a"
```

cpu (approved):
@tob-scott-a Is this still something you would be interested in seeing land?

It seems pretty reasonable to me (with the caveat that I have no experience with the PHP or composer ecosystems).

Perhaps once further progress is made on CI we could think of a way to smoketest the composer distribution.

tob-scott-a:
@cpu Yes, this will be a low-friction way to ensure PHP cryptography libraries, such as [PHP-ECC](https://github.com/paragonie/easy-ecc) and [phpseclib](https://github.com/phpseclib/phpseclib), can consume the latest Wycheproof tests for the algorithms they implement.

cpu:
> The composer.json file lets us pull the repo in as a dependency and never need any maintenance or ops overhead.

I'm convinced there's value to PHP users in having this, I'm just a bit worried it's going to scale poorly when users from other ecosystems want the same thing, or as the inevitable churn upstream creeps in and we have to adapt our side to match.

> In order for other people to wrap the wycheproof repository, they need to muck with gitsubmodules and the like, which is a bit of a headache.

That's fair, but it's also going to be a bit of a headache to keep up with packagist.org and the composer ecosystem from this side.

I'd still like to hear Filippo or someone else weigh in with their thoughts, but ignoring that for the moment maybe the solution here is to ask for help? :-) If we merged this could I count on your best-effort participation over time to keep it working or is this more of a one-and-done thing from your perspective?

cpu:
> I'd still like to hear Filippo or someone else weigh in with their thoughts

Chatting with Filippo on this subject we're :+1: on merging. I'm going to try to pull out some guidance for CONTRIBUTING.md for the future but the main points are:

* Like this branch, for future ecosystems we want to focus on solutions that make it easy to consume the JSON data and _not_ on providing deser. helpers/in-memory representations of the JSON data (e.g. annotated Go/Rust structs would be out-of-scope and should be built on top of the JSON)
* We're going to lean on the folks proposing support for their ecosystem of choice to help us maintain the machinery going forward. You've listed yourself as the point of contact in the composer metadata so I think we're squared away there.

cpu:
Thank you! 

If for whatever reason exposing the files from another repo ends up being cumbersome I think we're OK with restoring the very thin PHP bits you had before. I'm likely going to add a small `.go` file that does similar for that ecosystem using the [`embed` package](https://pkg.go.dev/embed) in the coming days (we already have a `go.mod` from the schema linting tooling). Mainly we want to avoid having to keep more involved code in-sync with the vector structure over time.

composer.json
```diff
@@ -0,0 +1,22 @@
...
+        }
+    ],
+    "type": "library",
+    "keywords": ["wycheproof", "testing", "cryptography", "crypto"],
+    "license": "Apache-2.0",
+    "autoload": {
+        "psr-4": {
+            "C2sp\\Wycheproof\\": "php"
+        }
+    },
```
cpu, line 18: @tob-scott-a just before I merge, should this part of the config be deleted to reflect 34982b8ac8f19cca05405bb7a4ba2b837f2bf117 ? 

(Also if you wouldn't mind adding a trailing newline to this file that would be great)

cpu (approved):
:rocket:

tob-scott-a:
`tob-scott-a` is my Packagist username. Don't worry about that GitHub Hook error message, it tends to go away after you press Update once.

cpu:
Hmm. I'm still seeing the error after hitting "update" on the packagist page. 

It links to [this doc page](https://packagist.org/about#how-to-update-packages) that has more detailed instructions in case it's helpful.

I can also add you as a maintainer on the packagist side if you confirm your username in case that helps (?)

PR #110, opened 2024-03-28
Test vectors for Kyber and ML-KEM.

sophieschmieg:
Specifically, for round 3 and for the NIST Draft standard, as well as the discussed potential modification of the draft standard that does silently reduce instead of failing on unreduced vectors:
* The vectors of the round 3 submission package
* Vectors where public or private keys are not reduced mod q
* Vectors where the various parts of Kyber are too short or too long
* Edge cases where the secret and/or the error are zero
* Vectors where the ciphertext is random bytes
* Bit flips in ciphertext
* message all zero/all 0xff
* Values of rho where SHAKE expands more than usual and read up to 591 bytes.
* Values of rho where the matrix has relatively large values (maximizing the sum of all entries)
* Values of rho where the matrix contains an unusual amount of zeroes in NTT form (I found a seed with 3 zeroes mod prime factor of (3329), and a number of seeds with 2 zeroes)
* Values of rho for which the matrix fails to be invertible mod (3329), which is otherwise a property that a random matrix is expected to have with high probability.


rben-dev:
Hi,

Thanks a lot for sharing these useful ML-KEM edge cases test vectors!

Are there any updates planned for the finalized FIPS203 ML-KEM release from August 2024 which slightly differs from the previous NIST draft? (namely the addition of domain separation for K-PKE.KeyGen and the swapped indices for the matrix access).

Thanks in advance,
Regards,

cpu:
> Specifically, for round 3 and for the NIST Draft standard

@sophieschmieg Would you be willing to regenerate these based on the finalized FIPS 203 spec? I would be very keen to see these land in-tree ASAP and I think that's the primary blocker.

FiloSottile:
@sophieschmieg happy to do the leg work of reformatting these in a Wycheproof format as promised if you can update them to the final FIPS :) I *think* we can't do that easily on our side because some seeds will need to be re-bruteforced?

PR #130, opened 2025-03-18
expand source metadata to facilitate external contributions

cpu:
Previously all Wycheproof vectors originated from the same source, Google, and so tracking a generator version in a `generatorVersion` field of the various schemas was sufficient provenance.

As the project expands into a community format and looks to encourage external parties/projects to add vectors it's beneficial to have more robust metadata about where the vectors came from.

Additionally, three schemas not referenced by any test vectors are removed:

* ec_point_test_schema.json
* ec_public_key_verify_schema.json
* rsa_key_test_schema.json


FiloSottile:
Having the `source` at the top level makes it impossible to add new vectors to a file from a new source, which I hope will be a common occurrence. We could say "make a new file for the same test type with a different source" but that makes it quite a bit more annoying to consume them and I imagine some consumers might just not pick up the new files, which is unfortunate. I was imagining `source` would be at the `testGroup[]` level. This does mean we need to remove the top-level `generatorVersion`. I can't imagine that causing severe disruption.

I am not sure how `name` and `generator` would differ, which to me suggests they are redundant. Also, I am not sure what semantics and responsibilities `maintainers` encodes. We can add that later if we decide to make maintenance more structured. `name` and `version` are probably enough.

Something I am not sure about is how to assign/reconcile the `tcId`s. We can learn as we add more sources.

cpu:
Thanks for the feedback :rocket: 

> I was imagining source would be at the testGroup[] level. 

That sounds reasonable, I will adjust :+1: 

> This does mean we need to remove the top-level generatorVersion. I can't imagine that causing severe disruption.

Agreed.

> I am not sure how name and generator would differ, which to me suggests they are redundant.

I was probably overthinking things here. I was imagining `name` would identify a repo/group and within that repo/group there might be multiple generators.

> Also, I am not sure what semantics and responsibilities maintainers encodes. We can add that later if we decide to make maintenance more structured. name and version are probably enough.

Mostly I just wanted to have some sense of who we should try to contact for topics related to the vectors. E.g. if someone reports an error, or if we're making breaking changes. It'd be handy to not have to dig back through `git blame` for that (and it may change over time).

I'll remove it for now - I think I agree it's better to start small & see how it goes.

cpu:
> cpu marked this pull request as ready for review now

I simplified the source schema as requested, and moved it to the test group elements. I went ahead and updated each of the schemas and the vector files that reference them to remove the legacy generator field, and to add a new source object per-testgroup that matches the version that was in the legacy generator field.

There were 3 schemas that weren't referenced by any vectors, so I removed them:

* ec_point_test_schema.json
* ec_public_key_verify_schema.json
* rsa_key_test_schema.json

FiloSottile (approved):

PR #129, opened 2025-03-17
Initial linting/validation of JSON vectors and associated schemas

cpu:
:wave: Here's the start on some CI infrastructure and vector/schema validation. I suspect we can come up with lots of other interesting properties to lint for, but for now I'm focused on the JSON schemas. This PR is a good place to start, but isn't an end-state and comes with several TODOs for follow-up.

### tools: add vector linting tool
This commit initializes a Go module for the repo, adding a single command under `tools/vectorlint`.

The new `vectorlint` tool can be provided a directory of schemas, and one or more directories of test vectors. It will process all vectors to check that:

* the vector JSON is valid JSON content
* the vector JSON references a schema JSON file that exists
* the schema JSON compiles
* the vector JSON is valid according to its referenced schema

This initial work has some important limitations to be addressed with follow-up PRs:

1. There are 7 schemas referenced by vector files in `testvectors_v1/` that don't exist. For now these are tracked in a `missingSchemas` map and the tool ignores vectors that ref these. We should add the missing schemas and remove the `missingSchemas` entries. 

2. The schemas use a number of custom formats. To get started I've implemented the format validation as a no-op. We should add validation logic for each custom format and remove the no-op validation.

3. The schema's don't include `"additionalProperties":false`, so unknown fields are allowed.

### ci: add simple GitHub actions vectorlint CI

Ensures the Go code remains nicely formatted, builds without error, and that the `vectorlint` tool finds no issues.

This task is run for pushes, PRs, and on a cron schedule to prevent any unexpected regressions/issues.

Here's [an example passing run](https://github.com/cpu/wycheproof/actions/runs/13909922164/job/38921691561) from my fork.

### Example output:

<details>
<summary>Output from go run ./tools/vectorlint:</summary>

```
2025/03/17 17:07:57 reading schemas from "schemas"
2025/03/17 17:07:57 reading vectors from ["testvectors_v1" "testvectors"]
2025/03/17 17:07:57 ✅ "testvectors_v1/a128cbc_hs256_test.json": validates with "aead_test_schema.json"
2025/03/17 17:07:57 ✅ "testvectors_v1/a192cbc_hs384_test.json": validates with "aead_test_schema.json"
2025/03/17 17:07:57 ✅ "testvectors_v1/a256cbc_hs512_test.json": validates with "aead_test_schema.json"
2025/03/17 17:07:57 ✅ "testvectors_v1/aead_aes_siv_cmac_test.json": validates with "aead_test_schema.json"
[...]

PR #140, opened 2025-04-01
Add additionalProperties=false throughout schemas, fixup findings

cpu:
This branch continues from https://github.com/C2SP/wycheproof/pull/137 and https://github.com/C2SP/wycheproof/pull/129 by tightening up the existing JSON schemas to better match the actual test data.

The existing schemas had two issues that made them much less effective at validating the data JSON than they would be otherwise:

1. They (_with a few exceptions_) didn't specify `"additionalProperties": false` when spec'ing object properties. Unfortunately this meant that property keys that weren't specified were allowed.
2. They didn't specify required fields. As a result, data missing specified properties wasn't an issue.

The combination of these two effects meant almost all JSON would "validate" against the schema as long as there were no conflicts with property keys that _were_ defined. This in turn masked a number of errors in the schemas when verifying both the v0 and v1 vectors.

This branch resolves the first issue by adding `"additionalProperties": false` throughout and fixing the issues that fell out. These largely fell into a few buckets:

* v0 vector data that was missing fields in the schema (e.g. `curve`, `encoding`).
* v0 vector data that was using a different casing for field names (e.g. `keySize` vs `keysize`).
* v1 vector data that specified the v0 schema, but used different property names/shapes.

For the first two cases I fixed the issues in the existing schema files. For the last case, I "forked" the v0 schema into a "v1" schema to use alongside, switched the v1 vectors to reference the forked schema, and fixed the schema to match the v1 vector data.


FiloSottile:
LGTM. Deleted `.gitignore`, IDE configs can go to `.git/info/exclude` locally.

A good next step might be to hoist some definitions into `common.json`.

cpu:
Thanks!

> A good next step might be to hoist some definitions into common.json.

Agreed - tracking that in https://github.com/C2SP/wycheproof/issues/141

PR #142, opened 2025-04-03
Remove a couple leftover BUILD.bazel files

FiloSottile:


cpu (approved):
Ooops, thanks for catching/fixing :+1:

PR #145, opened 2025-05-01
change some rsa oaep tests to acceptable

reaperhulk:
NIST SP800-56Br2 Section 7.1.2.1 specifies that when decrypting a ciphertext outside the range 1 < ct < (n-1) it should be rejected. The test cases updated in this commit had a ct integer value of 1 or 0, which violate this requirement.

This also adds a new flag `SmallIntegerCiphertext` to make it easier for consumers to skip on this.


cpu (approved):
LGTM.

Would you mind squashing everything into one commit? Alternatively we could squash-merge on our end. Either works for me :-)

reaperhulk:
I can squash it, one moment 😄

alex (approved):

PR #152, opened 2025-07-21
Refinment of project scope & documentation updates

cpu:
There's a large number of important downstream cryptography projects relying on Wycheproof for test vector data. Let's optimize our documentation/mission/priorities for increasing the value of those vectors through streamlining external contributions and continued community maintenance, while also advocating for maintainers to integrate our test vectors in their own CI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,16 @@ At the time of writing, the following `testvectors_v1` files are missing schemas
 
 Contribution of schemas for the above vectors would be most welcome.
 
+### Is there additional documentation about test vectors?
+
+Some legacy documentation for [files](doc/files.md), [formats](doc/formats.md) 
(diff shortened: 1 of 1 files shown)


J08nY:
I believe the removal of docs like [fd3a5ab](https://github.com/C2SP/wycheproof/pull/152/commits/fd3a5aba95b76658618513d04a90dd54f6822583) is not a good idea. Having just a set of pure test-vectors is fine, but it leaves all the responsibility into figuring out what a test-vector failure means on the user of the test-vectors. Some of these attacks/implications are tricky to workout and documentation on the origin/background/implications of the test-vectors is very much appreciated.

cpu:
@J08nY thanks for the feedback. With more reflection I agree with you and will undo those changes.

PR #153, opened 2025-07-29
docs: improve signature malleability wording

josibake:
Apologies for the churn on this, but while [reviewing](https://github.com/bitcoin-core/secp256k1/pull/1711#discussion_r2230819323) the downstream PR to integrate #150, I noticed this comment seems incorrect.

I believe it should read something like the following to indicate these test vectors are only for implementations that _do not_ tolerate signature malleability (corroborated by another reviewer [here](https://github.com/bitcoin-core/secp256k1/pull/1711#discussion_r2239970542))


cpu (approved):
This updated verbiage reads correctly to me as well. Thank you for following up on this upstream :-)

PR #161, opened 2025-08-20
Add ECDSA tests with non-minimally encoded DER tags

davidben:
DER tags are typically one byte, but when the tag number exceeds 30, they are encoded in a multi-byte tag form. The tag number bits of the first byte is set to all ones (i.e. 31) and then the actual tag number is encoded in the big-endian base128 encoding. (See X.690 (02/2021) 8.1.2.4.)

This encoding is only for larger tag numbers, so these should be rejected by an ECDSA signature verifier. This case was missing from the original Wycheproof tests.

Although this is a general DER issue, it makes sense to test this specifically in ECDSA test vectors because ECDSA is deep in the crypto layer of libraries, and so it is not uncommon for this to be implemented with slightly different parsers than the rest of the system.

These tests are tagged InvalidEncoding and not BerEncodedSignature because non-minimal integers is forbidden in BER too. X.690 (02/2021) 8.1.2 says that, for even BER, the single-octet encoding is the encoding for tags 0 to 30 and the multi-octet encoding is only for 31 and up.

I've added the tests to both testvectors and testvectors_v1 since it was easy, though hopefully that situation will settle out soon.

The tests were generated by the following script:
<details>
<summary>make_bad_tags.py</summary>

```
import json
import sys

SEQUENCE = 0x30
INTEGER = 0x02

def parse_elem(inp):
    # Parse a tiny subset of DER.
    tag, inp = inp[0], inp[1:]
    assert tag & 0x1f != 0x1f
    l, inp = inp[0], inp[1:]
    if l > 0x80:
        assert l == 0x81
        l, inp = inp[0], inp[1:]
        assert l >= 0x80
    body, inp = inp[:l], inp[l:]
    return tag, body, inp

def parse_sig(sig):
    tag, sig, rest = parse_elem(sig)
    assert tag == SEQUENCE
    assert len(rest) == 0
    tag, r, sig = parse_elem(sig)
    assert tag == INTEGER
    tag, s, sig = parse_elem(sig)
    assert tag == INTEGER
    assert len(sig) == 0
    return r, s

def encode_der(tag, body, long_tag = False):
    ret = bytearray()
    if long_tag:
        ret.append(tag | 0x1f)
        ret.append(tag & 0x1f)
    else:
        ret.append(tag)
    if len(body) < 0x80:
        ret.append(len(body))
    else:
        assert len(body) <= 0xff
        ret.append(0x81)
        ret.append(len(body))
    ret.extend(body)
    return bytes(ret)

for arg in sys.argv[1:]:
    with open(arg) as f:
        data = json.load(f)

    group = data["testGroups"][0]
    if group["type"] != "EcdsaVerify":
        print(f"{arg} is not an EcdsaVerify file. Skipping.")
        continue

    test = group["tests"][0]
    if test["result"] != "valid":
        raise ValueError("could not find sample signature")
    msg = bytes.fromhex(test["msg"])
    sig = bytes.fromhex(test["sig"])
    r, s = parse_sig(sig)

    bad_seq = encode_der(SEQUENCE, encode_der(INTEGER, r) + encode_der(INTEGER, s), long_tag=True)
    bad_r = encode_der(SEQUENCE, encode_der(INTEGER, r, long_tag=True) + encode_der(INTEGER, s))
    bad_s = encode_der(SEQUENCE, encode_der(INTEGER, r) + encode_der(INTEGER, s, long_tag=True))

    num_tests = data["numberOfTests"]
    new_group = {
        "type": "EcdsaVerify",
        "source": {
            "name" : "github/davidben",
            "version" : "0.1"
        },
    }
    for key in ( "key", "keyDer", "keyPem", "publicKey", "publicKeyDer", "publicKeyPem", "sha"):
        if key in group:
            new_group[key] = group[key]
    new_group["tests"] = []

    def append_test(comment, sig):
        global num_tests
        num_tests += 1
        new_group["tests"].append({
            "tcId": num_tests,
            "comment": comment,
            "flags": ["InvalidEncoding"],
            "msg": msg.hex(),
            "sig": sig.hex(),
            "result": "invalid",
        })

    append_test("signature with non-minimal SEQUENCE tag", bad_seq)
    append_test("signature with non-minimal INTEGER tag on r", bad_r)
    append_test("signature with non-minimal INTEGER tag on s", bad_s)

    data["testGroups"].append(new_group)
    data["numberOfTests"] = num_tests
    with open(arg, "w") as f:
        json.dump(data, f, indent=2, separators=(",", " : "))
        f.write("\n")
    print(f"Added new tests to {arg}")
```

</summary>

### DIFF
--- a/testvectors/ecdsa_brainpoolP224r1_sha224_test.json
+++ b/testvectors/ecdsa_brainpoolP224r1_sha224_test.json
@@ -1,6 +1,6 @@
 {
   "algorithm" : "ECDSA",
-  "numberOfTests" : 359,
+  "numberOfTests" : 362,
   "header" : [
     "Test vectors of type EcdsaVerify are meant for the verification",
     "of ASN encoded ECDSA signatures."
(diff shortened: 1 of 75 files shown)


testvectors/ecdsa_brainpoolP224r1_sha224_test.json
```diff
@@ -4587,6 +4587,56 @@
           ]
         }
       ]
+    },
+    {
+      "type" : "EcdsaVerify",
+      "source" : {
```
davidben, line 4593: Is this the right convention for `source`? Or should I be further subdividing it like `github/davidben/non-minimal-tag`? Not sure what the intention is with how sources are to be managed, and whether it's envisioned to feed into tooling.

cpu: My vision for the source field is that it should eventually feed into tooling to allow splicing in updates/replacements of a subset of the .json data based on matching the source/version.

If you think you might add more ECDSA test cases in the future for a different class of issue, and would want to update them separately from these, it might make sense to be more specific.

cpu (approved):
These are great, thank you for the contribution.

testvectors/ecdsa_brainpoolP224r1_sha224_test.json
davidben: Sounds good! Not sure, but doesn't hurt to make it more specific.

cpu: > (edit: and we should probably capture some advice on this topic in CONTRIBUTING.md too)

I took a crack at porting some of the outcome of this discussion into docs as part of  https://github.com/C2SP/wycheproof/pull/169

PR #172, opened 2025-09-01
Split ML-KEM tests into separate files by parameter

### LINKED ISSUE #166: ML-DSA and ML-KEM handle parameters differently
ML-DSA and ML-KEM are each a family of three algorithms. The ML-DSA test vectors work by splitting them up into separate files, while the ML-KEM test vectors have one big file and a parameterSet key.

It would be nice if we could be consistent one way or another. (If I had to pick, I like the ML-DSA pattern better. We don't support every ML-KEM variant and being able to only use the files that matter is nicer.)

davidben:
Fixes #166


davidben:
Yeah, I'm similarly dubious `numberOfTests` does anything useful.

Issue #171, opened 2025-09-01
RSA-PSS public keys are invalid DER


davidben:
(I found my old notes from when I first tried to import testvectors_v1 and realized there was a bug I never filed...)

The id-RSASSA-PSS public keys aren't encoded right. For example, looking at the first `publicKeyDer` in `testvectors_v1/rsa_pss_2048_sha1_mgf1_20_params_test.json`, here is the first public key:
```
-----BEGIN PUBLIC KEY-----
MIIBUzA+BgkqhkiG9w0BAQowMaALMAkGBSsOAwIaBQChGDAWBgkqhkiG9w0BAQgw
CQYFKw4DAhoFAKIDAgEUowMCAQEDggEPADCCAQoCggEBAL0xx6AmkdLZWH72qUb/
eIVEzK3UspiK1iCGeSpr+WyGFrStEzF9InC5AdD80diAy49S+4cwSlJYwRs43+ro
32cK7ufqHQ2d+OAOgIR+QeWYntQC1E54sw/vF7VnHTrb+GheTcIESZ7NGGPh1a/y
inz2bq3zH+ySNsEgrdE0UVIsZHyYMqZyzWTTKMHDIhg/RmHQm9pguN1fAyjaVCCC
FCSv2rsagMXRJ2OhsCOM2J0HQr/FC2ovy3AdgkIY+YJvT3iiOitapCrOfxdTdvts
vbK60pO6WD1NMca4+QKeRrE2iSSYVfUFdW4A4iWmpFoYdpvY0rOkrLnxwj0+UYgl
YeUCAwEAAQ==
-----END PUBLIC KEY-----
```

However, if we look at the decoding in [der2ascii](https://github.com/google/der-ascii):
```
      [0] {
        SEQUENCE {
          # sha1
          OBJECT_IDENTIFIER { 1.3.14.3.2.26 }
          NULL {}
        }
      }
      [1] {
        SEQUENCE {
          # mgf1
          OBJECT_IDENTIFIER { 1.2.840.113549.1.1.8 }
          SEQUENCE {
            # sha1
            OBJECT_IDENTIFIER { 1.3.14.3.2.26 }
            NULL {}
          }
        }
      }
      [2] {
        INTEGER { 20 }
      }
      [3] {
        INTEGER { 1 }
      }
```

[...]

PR #173, opened 2025-09-01
Remove redundant rsa_pss_*_shake*_params_test.json tests

davidben:
Each of these is identical to the corresponding non-params files. That is, rsa_pss_2048_shake128_params_test.json is identical to rsa_pss_2048_shake128_test.json, etc.

In the SHA versions, the difference is whether you encode as id-rsaEncryption or id-RSASSA-PSS (RFC 4055), with the latter capturing parameters that need to be matched against the sigalg parameters. That encoding wasn't defined for SHAKE until RFC 8692. The RFC 4055 encoding is absolutely horrid, so RFC 8692 (rightfully) departs from RFC 4055 entirely and just made new top-level OIDs.

Ideally the params version would just use the RFC 8692 encoding, but for now just remove the redundant ones. (I don't work on a stack that has implemented or has any plans to ever implement RFC 8692, so I don't want to fix those tests without checking.)

See https://github.com/C2SP/wycheproof/issues/171#issuecomment-3242849333


cpu (approved):
Good catch :+1: Thank you.

Doing a pass to compare other `testvectors_v1/` files by their digest to find duplicates also shows that there's a few `ecdsa_secp*_sha*_p*_test.json` files that are identical to their `ecdsa_secp*_webcrypto_test.json` counterparts. I'll rm the `_webcrypto_test.json` versions of those in a separate follow-up.

PR #181, opened 2025-09-15
Provide JSON schema for EcCurveTest

botovq:
The schema was mostly modeled after the one for the ML-KEM tests. It marks the generatorVersion as deprecated like in all other schemas and adds a source object to the TestGroup. Remove it from the list of missing schemas in the readme and the linter.

ec_prime_order_curve_test.json limits itself in scope to curves of prime order over prime fields. This doesn't seem entirely correct: First, its format is only really suitable for the short Weierstrass representation of a curve over a finite field. Second, that the order is prime is a desirable common feature of the curves it currently contains rather than a necessity. Perhaps the test should be renamed and then it could easily be extended in scope.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,6 @@ At the time of writing, the following `testvectors_v1` files are missing schemas
 
 * `testvectors_v1/aes_ff1_base*_test.json`
 * `testvectors_v1/aes_ff1_radix*_test.json`
-* `testvectors_v1/ec_prime_order_curves_test.json`
 * `testvectors_v1/ecdsa_secp256k1_sha256_bitcoin_test.json`
 * `testvectors_v1/pbes2_hmacsha*_aes_*_test.json`
 * `testvectors_v1/pbkdf2_hmacsha*_test.json`
--- a/schemas/ec_curve_test_schema.json
+++ b/schemas/ec_curve_test_schema.json
@@ -1,0 +1,160 @@
+{
+  "type": "object",
+  "properties": {
+    "algorithm": {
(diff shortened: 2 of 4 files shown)


davidben:
No strong opinions either way, but I guess it's because APIs for primitives built with Montgomery and Edwards curves (correctly) tend not to expose the parameterization, while we've been plagued with short-Weierstrass ECC APIs that expose all this stuff. (E.g. https://docs.oracle.com/javase/8/docs/api/java/security/spec/ECParameterSpec.html which not only exports the parameters, but also lets you construct custom curves.)

bleichenbacher-daniel:
The current schema is based on java.security.spec.ECParameterSpec.
This schema is indeed quite restrictive since, it does not even cover all Weierstrass curves in use. One example is BLS-377, which uses a group with a large cofactor (i.e. 0x170b5d44300000000000000000000000) and hence cannot be represented as an int.
The main purpose of these test vectors is simply to ensure that there are no confusion when different names for the same parameters are being used, and even more so when the same name is being used for different parameters (e.g. BN254 is such a case). 

The plan is to indeed extend the project to more curves. Currently I'm using the following curves: 
https://github.com/bleichenbacher-daniel/Rooterberg/blob/main/doc/ec_curves.md
BTW, there are quire a few groups with large cofactors. Since the edge cases for arithmetic errors that I construct are typically not in the group it is difficult to use them in tests against primitives such as ECDSA or ECDH. This means that additional tests underlying operations such as point multiplication are useful.

Issue #162, opened 2025-08-20
Use a more common JSON formatting


davidben:
When writing some test vectors, I noticed that the JSON formatting is a bit uneven. It looks like the original Wycheproof test put a space _before_ the `:` in JSON dictionaries, which is a bit uncommon. I see we've already had other test vectors, such as ML-DSA, which use the more conventional pattern: only a space _after_ the `:`.

I propose we use a uniform encoding, and use one without the extra space. (I.e. we run a tool to rewrite all the old vectors.) The value of a more common encoding is that contributors can more easily add test vectors without a ton of diff noise. This came up when I was putting together https://github.com/C2SP/wycheproof/pull/161 and had to figure out how to reproduce the original formatting. It happens that Python has a `separators` parameter to reproduce it but other APIs ([JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify), [Go](https://pkg.go.dev/encoding/json#MarshalIndent)) only control the indent . We should pick a convention that the most tools can reproduce easily.

(Even better would be if there were some tool for folding a contributor's test vector set into the combined file, and then the formatting would be whatever that tool thinks it is, but either way I think we would be better served by a more conventional formatting.)

cpu:
Thanks for opening this issue.

> I propose we use a uniform encoding, and use one without the extra space. (I.e. we run a tool to rewrite all the old vectors.) 

+1 - this has been on my mind as well. I think it makes sense to do a pass over the existing data and then enforce consistency going forward in CI with the same tool.

> Even better would be if there were some tool for folding a contributor's test vector set into the combined file

Agreed, @FiloSottile and I discussed a similar idea in an earlier conversation. I think that's a place where I'd like to invest some time/energy after #151 (though if someone else is keen to get started before I can get to it I'm happy to support their effort!)

davidben:
Happy to put up a PR if it would help. Mostly I didn't want to collide with the work you're doing in #151.

davidben:
Sure! Although in a similar unprecedented software engineer phenomenon, I realized that canonicalizing the indent with standard JSON implementations very quickly bumps into how to canonicalize the field order. On the one hand, sorting keys is very easy and something every tool can do. On the other hand, it is a bit sad to not put `tcId` first, etc.

[...]

PR #183, opened 2025-10-02
Reformat JSON test vector files, verify formatting in CI

cpu:
Adds a simple `tools/reformat_json.py` Python script to ensure consistent and uniform formatting, and to switch away from the less conventional use of " : " separators. We use `ensure_ascii=False` to preserve unicode content in comments (e.g. `2⁻¹⁶` in `mlkem_512_test.json`). CI is updated to run this script and fail if a diff is produced to prevent regression.

The script avoids formatting `aes_ff1_*.json` files for the reasons [outlined in this comment](https://github.com/C2SP/wycheproof/issues/162#issuecomment-3247171822). The delimiters in this file are fixed manually with `sed` in a separate commit, and we'll have to be careful to avoid regressions since it's excluded from CI checks. It seems unlikely these files will change anytime soon, so let's not get hung up on this while we can do better for the majority of files.

Python is used here because unfortunately Go makes it quite difficult to accomplish the same thing without reordering fields. I'm personally of the mindset that sorting the field order would make this easier and our primary consumers are machines and not humans, but an alpha sort has the disadvantage of moving fields like `tcId` that are currently up-front into a more or less random position. For now let's avoid this complication and use Python.

Updates https://github.com/C2SP/wycheproof/issues/162


davidben (approved):
LGTM

Verified the formatting by spot-checking a few and then confirming that `json.dumps(json.load(...))` before and after this PR produced the same output. (Am assuming that the GHA workflow will confirm that the formatting itself matches expectations.)

botovq (approved):

Issue #184, opened 2025-10-07
SPKI and PKCS8 for ML-KEM and ML-DSA tests


davidben:
Filing this to remember to make a note of it. Unless someone beats me to it, I'll probably put something together, at least for the seed-bases ones, as part of other work.

(I won't bother with expandedKey one as those just generally are bad ideas. The IETF fumbled the ball at the last minute and we ended up with three different PKCS8 encodings, two of which should never be used.)

tarcieri:
Was this addressed by #186?

cpu:
I thought there were still gaps for ML-KEM

Arpan0995:
Is the ML-KEM side of this still open? As of today none of the ML-KEM files carry a DER field: `mlkem_*_encaps_test`, `mlkem_*_keygen_seed_test`, `mlkem_*_semi_expanded_decaps_test` and `mlkem_*_test` have no `publicKeyDer` or `privateKeyPkcs8` equivalent, so the gap @cpu mentioned looks like it is still there. #186 covered ML-DSA only.

If nobody has it in flight I would be glad to put it together. One structural question first, because ML-KEM does not fit the #186 shape directly.

In the ML-DSA files the key lives on the test group, so #186 could add `publicKeyDer` and `privateKeyPkcs8` once per group. In the ML-KEM files `ek`, `dk` and `seed` are per test case, and there are 1701 of them across the twelve files: encaps is 261/265/269 for the three parameter sets, the main file and keygen_seed are 193 and 100 each, decaps is 9. So a per-case DER field is a much bigger change than the ML-DSA one was. Three ways to go:

1. Add `ekDer` and `dkPkcs8` per test case, consistent with where the raw keys already are, at the cost of a large size increase across the existing files.
2. Add them only in `keygen_seed`, which is the file whose subject is the key material anyway, and leave encaps/decaps untouched.
3. Keep them out of the existing files entirely and add a separate key-encoding file per parameter set.

I would lean towards 2, but this is your call on file layout rather than mine.

Two details worth settling at the same time:

The ML-KEM seed is 64 bytes (`d || z`), not 32 as for ML-DSA, so "the seed form" means something different here and the schema description should probably say so explicitly rather than reuse the ML-DSA wording.

[...]

PR #186, opened 2025-10-18
Add SPKI and PKCS8 inputs for ML-DSA

davidben:
Part of #184. I haven't done ML-KEM yet, but it should be fairly easy to extend the script.

Also fix a typo in mldsa_verify_schema.json. This went unnoticed because both missing and extra fields are allowed.

<details><summary>Generated with this script:</summary>

```
import json
import sys

OIDS = {
    "ML-DSA-44": b"\x60\x86\x48\x01\x65\x03\x04\x03\x11",
    "ML-DSA-65": b"\x60\x86\x48\x01\x65\x03\x04\x03\x12",
    "ML-DSA-87": b"\x60\x86\x48\x01\x65\x03\x04\x03\x13",
}

SEQUENCE = 0x30
INTEGER = 0x02
BIT_STRING = 0x03
OCTET_STRING = 0x04
OBJECT_IDENTIFIER = 0x06

CONTEXT_SPECIFIC = 0x80
CONSTRUCTED = 0x20

def encode_der(tag, body):
    ret = bytearray()
    ret.append(tag)
    if len(body) < 0x80:
        ret.append(len(body))
    elif len(body) <= 0xff:
        ret.append(0x81)
        ret.append(len(body))
    elif len(body) <= 0xffff:
        ret.append(0x82)
        ret.append(len(body) >> 8)
        ret.append(len(body) & 0xff)
    ret.extend(body)
    return bytes(ret)

def insert_after(d, pos, new_key, new_value):
    # This is silly. Maybe we can automatically order fields by the schema?
    # Or just alphabetize...
    if new_key in d:
        d[new_key] = new_value
        return
    assert pos in d
    items = list(d.items())
    d.clear()
    for k, v in items:
        d[k] = v
        if k == pos:
            d[new_key] = new_value

def algorithm_identifier_no_params(oid):
    return encode_der(
        SEQUENCE,
        encode_der(OBJECT_IDENTIFIER, oid)
    )

def fix_mldsa_verify(alg, g):
    key = bytes.fromhex(g["publicKey"])
    # This would just be adding a fixed prefix, except that some inputs
    # intentionally have the wrong length.
    spki = encode_der(
        SEQUENCE,
        algorithm_identifier_no_params(OIDS[alg]) +
        encode_der(
            BIT_STRING,
            b"\x00" + key,
        )
    )
    insert_after(g, "publicKey", "publicKeyDer", spki.hex())

def fix_mldsa_sign(alg, g):
    seed = bytes.fromhex(g["privateSeed"])
    pkcs8 = encode_der(
        SEQUENCE,
        encode_der(INTEGER, b"\x00") +
        algorithm_identifier_no_params(OIDS[alg]) +
        encode_der(
            OCTET_STRING,
            encode_der(CONTEXT_SPECIFIC | 0, seed)
        )
    )
    insert_after(g, "privateSeed", "privateKeyPkcs8", pkcs8.hex())

def fix(path):
    with open(path) as f:
        j = json.load(f)
    for g in j["testGroups"]:
        if j["schema"] == "mldsa_verify_schema.json" and g["type"] == "MlDsaVerify":
            fix_mldsa_verify(j["algorithm"], g)
        elif j["schema"] == "mldsa_sign_seed_schema.json" and g["type"] == "MlDsaSign":
            fix_mldsa_sign(j["algorithm"], g)
    with open(path, "w") as f:
        json.dump(j, f, indent=2, separators=(",", ": "), ensure_ascii=False)
        f.write("\n")

for arg in sys.argv[1:]:
    fix(arg)
```

</details>

### DIFF
--- a/schemas/mldsa_verify_schema.json
+++ b/schemas/mldsa_verify_schema.json
@@ -26,7 +26,9 @@
         "source": {
           "$ref": "common.json#/definitions/Source"
         }
-      }
+      },
+      "additionalProperties": false,
+      "required": ["type", "publicKey", "publicKeyDer", "tests"]
     },
     "MlDsaVerifyTestVector": {
       "type": "object",
@@ -66,13 +68,20 @@
           },
           "description": "A list of flags"
         }
-      }
+      },
+      "additionalProperties": false,
+      "required": ["tcId", "comment", "msg", "sig", "flags"]
     }
   },
   "properties": {
     "algorithm": {
       "type": "string",
       "description": "the primitive tested in the test file"
+    },
+    "generatorVersion": {
+      "type": "string",
+      "description": "DEPRECATED: prefer \"source\" property in test group",
+      "deprecated": true
     },
     "header": {
       "type": "array",
@@ -97,5 +106,7 @@
         "$ref": "#/definitions/MlDsaVerifyTestGroup"
       }
     }
-  }
+  },
+  "additionalProperties": false,
+  "required": ["algorithm", "header", "notes", "numberOfTests", "schema", "testGroups"]
 }


schemas/mldsa_verify_schema.json
```diff
@@ -7,11 +7,16 @@
         "type": {
           "enum": ["MlDsaVerify"]
         },
-        "public_key": {
+        "publicKey": {
```
cpu: > Also fix a typo in mldsa_verify_schema.json. This went unnoticed because both missing and extra fields are allowed.

Argh :face_exhaling: Thanks for fixing. Such a frustrating default for a schema language!

I thought we managed to get `"additionalProperties": false` in all the schemas but this one must have been overlooked. A quick `grep -rL '"additionalProperties"' schemas/` on `main` only flagged `mldsa_verify_schema.json`.

Would you be interested in sprinkling `additionalProperties` & some `"required"` fields into this schema's definitions as part of this PR, or would you prefer I do it as a follow-up?

botovq (approved):
The schema changes also read fine.

PR #198, opened 2025-12-01
x25519/x448: audit and fixup Twist flag test cases

cpu:
This is an extension of https://github.com/C2SP/wycheproof/pull/196 with a few differences. Please check my work :-)

1. The `x25519_test.json` test case (tcId 101) that @XoifaiI flagged was also present in the `x25519_pem_test.json` and `x25519_jwk_test.json` variations. I amended their commit to fix these instances in the same manner as was done in `x25519_test.json` (Another point in favour of having less formats...)
2. I believe both the x25519 and x448 test cases had some acceptable result test cases that were missing the "Twist" flag.
3. I believe the x25519 test cases had some acceptable result test cases that had the "Twist" flag, but that weren't twist points.

I've included [the program](https://github.com/cpu/wycheproof/blob/cpu-do-the-twist/tools/twistcheck/twistcheck.go) I wrote to lint these files, and integrated it with CI. This is mainly so others can check my work. I could be convinced it's not worth including the tool + CI integration. Please chime in if you have opinions here.

### DIFF
--- a/tools/twistcheck/twistcheck.go
+++ b/tools/twistcheck/twistcheck.go
@@ -1,0 +1,241 @@
+// twistcheck verifies that X25519 and X448 test vectors with twist points are
+// correctly marked with the Twist flag. Additionally, the result must be
+// "acceptable" since specific implementations may choose to reject twist points.
+package main
+
+import (
+	"crypto/x509/pkix"
+	"encoding/asn1"
+	"encoding/base64"
+	"encoding/hex"
+	"encoding/json"
+	"encoding/pem"
+	"flag"
+	"fmt"
+	"log"
+	"math/big"
+	"os"
+	"slices"
+	"strings"
+
+	"filippo.io/edwards25519/field"
+)
+
+var (
+	vectorFile = flag.String("vectors", "", "path to test vector file")
+)
+
+func main() {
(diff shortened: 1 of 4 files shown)


XoifaiI:
Can confirm all the tests (including x25519) now pass for [this](https://github.com/daily3014/rbx-cryptography) now
image

tools/twistcheck/twistcheck.go
```diff
@@ -0,0 +1,248 @@
...
+
+// isPointOnTwist448 checks if a point is on the twist of Curve448.
+// A point is on the twist if x³ + 156326x² + x is NOT a quadratic residue mod p.
+func isPointOnTwist448(publicKey []byte) (bool, error) {
+	// Curve448 field: p = 2^448 - 2^224 - 1
+	p := new(big.Int).Lsh(big.NewInt(1), 448)
+	p.Sub(p, new(big.Int).Lsh(big.NewInt(1), 224))
+	p.Sub(p, big.NewInt(1))
+
+	x := new(big.Int).SetBytes(bigEndian(publicKey))
```
rolandshoemaker: Incredibly not important, but you can use `slices.Reverse`

PR #202, opened 2025-12-11
Add ML-KEM decapsulation validation test vectors

sgmenda:
For https://github.com/aws/aws-lc/pull/2891, I made some simple vectors to validate full decaps keys. I needed a new schema because the existing `mlkem_test_schema.json` only supports seed keys, and not full keys: https://github.com/C2SP/wycheproof/blob/main/schemas/mlkem_test_schema.json#L80-L84


cpu:
Thanks!

PR #211, opened 2026-01-30
Add some more ML-DSA cases

chrisfenner:
This PR introduces 100 (for each of ML-DSA-44, -65, and -87) happy-path ML-DSA signing cases that include:

- Private key as a seed (each test case unique)
- Mu
- Randomness for hedged signing

This requires some new schema work. For the new test cases I have introduced a new `CompletePureMlDsaSignTestVector` to the common ML-DSA schema file.

These are taken from https://github.com/post-quantum-cryptography/KAT with a small modification by myself (https://github.com/chrisfenner/mldsa-test-vectors) to compute and verify Mu values.

Apologies if I've messed up any schema or naming conventions, I've made my best attempt to conform to the existing patterns.


chrisfenner:
> Exactly. I don't think my remaining time budget for Wycheproof this week will fit both but I will get to it "soon" ™️ :-)

Understood! I'm in no particular rush. I am able to use these in their current form but was asked to contribute these "upstream" in case it helps others. Let me know how I can be of help!

FiloSottile:
FWIW, I seem to remember that https://github.com/post-quantum-cryptography/KAT vectors are random, not targeting different edge cases each. If that's the case, I think having one for each parameter set and inputs combination is enough. In other words, if two vectors can't be expected to ever separately fail, there's no need to have two.

Looking only at the PR description, I think we already have Mu and seeds in the current schemas, so maybe we just need to add an optional randomness field to the existing schema and one vector per parameter set (-44, -65, and -87) with non-missing randomness?

chrisfenner:
Filippo got to this before I was able to, in #248. I'll close this PR in favor of letting the team look at that one. Thanks all!

PR #232, opened 2026-04-13
schemas: add pbkdf test schema

cpu:
This was one of the schemas that was referenced by test vector files, but never provided. Let's add a JSON schema and remove the exception from `vectorlint.go`.

### DIFF
--- a/tools/vectorlint/main.go
+++ b/tools/vectorlint/main.go
@@ -78,9 +78,6 @@ var (
 		// testvectors_v1/pbes2_hmacsha*_aes_*_test.json:
 		"pbe_test_schema.json": true,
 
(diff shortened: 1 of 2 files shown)


tools/vectorlint/main.go
```diff
@@ -78,9 +78,6 @@ var (
 		// testvectors_v1/pbes2_hmacsha*_aes_*_test.json:
 		"pbe_test_schema.json": true,
 
-		// testvectors_v1/pbkdf2_hmacsha*_test.json:
-		"pbkdf_test_schema.json": true,
```
cpu, line 82: Specifically tackling this one to use downstream in https://go-review.googlesource.com/c/go/+/758861

FiloSottile (approved):

PR #235, opened 2026-04-18
mldsa: add ctx-length=255 boundary sign-seed vectors

tob-scott-a:
I tried to identify areas to strengthen the ML-KEM and ML-DSA test vectors using [vector-forge](https://github.com/trailofbits/skills/tree/main/plugins/trailmark/skills/vector-forge) Claude skill, which uses mutation testing and graph analysis (ideally with multiple implementations) to identify areas where bugs are not killed by the current test vector suite. (Or to generate a fresh suite for an algorithm that doesn't have good tests already.)

After a list of candidates was identified, I had Claude eliminate duplicated work in other PRs (e.g., #234, which kills the use_hint mutants in RustCrypto).

The only ML-KEM mutation-testable gap found in RustCrypto was [in compress.rs](https://github.com/RustCrypto/KEMs/blob/584f5815a0dfa759a78d9f279d24113d97518bed/ml-kem/src/compress.rs#L18) and requires a targeted-search tool that generates a ciphertext whose decapsulated shared secret depends on decompression rounding direction. (The escaped mutant in question affects the `POW2_HALF` value.)

This isn't the most important addition to the test suite (and the ML-KEM one might require a bit of luck), but I thought I'd toss it over in case it's helpful to exercise the boundary conditions of ctx length.


Panchajanya1999:
works fine here, should merge

cpu:
> The existing 255-byte context test in each file (tcId 4) carries a precomputed mu and therefore verifies via Sign_mu / Verify_mu, bypassing the ctx-length check in Sign (FIPS 204 §5.2 line 3, pseudocode `if |ctx| > 255 return perp`). These new vectors exercise Sign directly with msg + ctx, so an implementation that mis-handles the boundary as `|ctx| >= 255` will reject a case the spec requires to succeed.
>
> Each vector still carries a precomputed mu so existing harnesses that only consume the mu path continue to match

I think we should remove the precomputed `mu` from the new tests. 
 
In the [WIP Go stdlib harness](https://go-review.googlesource.com/c/go/+/758781/4) we're using the presence of `mu` as a hint to use the external mu API (`SignExternalMuDeterministic`) vs computing it from msg/ctx (`SignDeterministic`). When we run these new vectors with a buggy boundary condition for the context they continue to pass because of the pre-computed `mu` pushing us towards the API that doesn't hash the context. I'm open to the notion that we should handle this differently in our harness but it seems like something others are likely to trip on as well and I'm not sure there's a lot of value in including `mu` for these new cases. 

WDYT?

cpu:
> works fine here

@Panchajanya1999 In the future this sort of comment would be more constructive with some additional information on what you tested & how. Otherwise it reads as noise.

FiloSottile:
> In the [WIP Go stdlib harness](https://go-review.googlesource.com/c/go/+/758781/4) we're using the presence of `mu` as a hint to use the external mu API (`SignExternalMuDeterministic`) vs computing it from msg/ctx (`SignDeterministic`).

Why wouldn't we do both? Good test harnessing always involves checking both vector <-> implementation correspondence and implementation-internal invariances. Sign(msg, ctx) == SignMu(mu) is one such invariant.

It'd be weird to exclude the mu from a test for which it can be validly computed. What if someone has a ComputeMu() function and they are using the vectors to test that?

cpu (approved):

cpu:
Since this has been open for a while without any additional feedback I'm going to merge it as-is. The Go harness has already been reworked to be more robust and doesn't have any issues with the new vectors.

PR #241, opened 2026-05-29
Fix aes_ff1_base85 alphabet errors

### LINKED ISSUE #239: aes_ff1_base85_test.json: alphabet has duplicate symbols
### Summary

In `testvectors_v1/aes_ff1_base85_test.json` (and the test groups it contains), every test group declares this `alphabet`:

```
0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrtstuvxyz!#$%&()*+_;<=>?@^_`{|}~
```

It is 85 characters long but contains **only 83 distinct characters** — it cannot be a valid radix-85 alphabet:

- the lowercase run reads `…opqrtstuvxyz` instead of `…opqrstuvwxyz` — `s` and `t` are transposed/duplicated (`…q t s t u v…`) and **`w` is dropped entirely**;
- `_` (underscore) appears **twice** (once in `…*+_;…`, once in `…@^_`{…`).

So `t` and `_` are each used for two different digit positions, and `w` has no digit at all. The digit↔symbol mapping is therefore ambiguous (does `_` mean digit 71 or 79?) and, under any reading, does not agree with the companion vectors.

### Impact

`aes_ff1_radix85_test.json` carries the same 1852 test cases as digit arrays (same `tcId` / `key` / `tweak` / `flags` / `result`), and the digit form is fine — the bug is purely in the symbol-alphabet rendering used by `aes_ff1_base85_test.json`. Cross-checking the two: under the natural left-to-right reading of the declared alphabet, **868 of the 1852** `msg`/`ct` strings in `aes_ff1_base85_test.json` do **not** decode back to the corresponding `aes_ff1_radix85_test.json` digit sequences.

Consequences for implementers:

- An FF1 implementation that uses a *correct* 85-distinct-symbol alphabet will fail these vectors.
- An implementation can only pass them by reproducing the typo (and even then must pick one of the two readings for `_`).
- Strict validators that reject non-injective alphabets reject this file outright.

### Suggested fix

Replace the alphabet with 85 distinct symbols — minimally: restore the lowercase run to `…opqrstuvwxyz` (insert `w`, fix the `s`/`t` order) and replace one of the two `_` with an unused graphic-ASCII symbol (e.g. `:`), giving:

```
0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz!#$%&()*+:;<=>?@^_`{|}~
```

(same first 54 characters as the buggy one; diverges only in the lowercase run and one symbol; 85 distinct characters)

— then re-render `msg`/`ct` for all 1852 test cases by mapping the `aes_ff1_radix85_test.json` digit values through the corrected alphabet. We did exactly this regeneration locally and round-trip-verified it: every corrected `msg`/`ct` decodes back to the radix-85 digits, **0 failures across all 1852 cases**. (Happy to share the regenerated file or a regeneration script.)

### History

This was first found against `google/wycheproof`, but a fix could not be landed at the time. The file carried over to `C2SP/wycheproof` unchanged.

### Environment

- `C2SP/wycheproof@master`, `testvectors_v1/aes_ff1_base85_test.json`, `generatorVersion: "0.9rc5"`
- Verified present and unfixed on 2026-05-11

cpu:
Implements the fixed vectors offered by @dspdon in https://github.com/C2SP/wycheproof/issues/239

* Corrects the base85 alphabet. Previously it was only 83 distinct characters: `s`/`t` were transposed, w was omitted, and `_` appeared twice.
* Standardizes the invalid-plaintext sentinels. Previously digit `-1` was rendered inconsistently as one of 10 different out-of-alphabet characters `(', ,, -, ., /, :, [, \, ], w)`, and digit 85 was rendered as `\x7f` (non-printable DEL). The updated vectors use `-` for both (which is not in the new alphabet, so the test cases remain invalid).

I did a quick sanity pass on the other `aes_ff1_*` testvectors and didn't spot any similar issues in their alphabets. Further review input welcome.

Resolves #239

### DIFF
--- a/testvectors_v1/aes_ff1_base85_test.json
+++ b/testvectors_v1/aes_ff1_base85_test.json
@@ -42,7 +42,7 @@
   "schema": "fpe_str_test_schema.json",
   "testGroups": [
     {
-      "alphabet": "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrtstuvxyz!#$%&()*+_;<=>?@^_`{|}~",
+      "alphabet": "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz!#$%&()*+:;<=>?@^_`{|}~",
       "keySize": 128,
       "msgSize": 0,
       "radix": 85,
@@ -63,7 +63,7 @@
       ]
     },
     {
-      "alphabet": "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrtstuvxyz!#$%&()*+_;<=>?@^_`{|}~",
+      "alphabet": "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz!#$%&()*+:;<=>?@^_`{|}~",
       "keySize": 128,
       "msgSize": 1,
       "radix": 85,
@@ -84,7 +84,7 @@
       ]
     },
     {
-      "alphabet": "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrtstuvxyz!#$%&()*+_;<=>?@^_`{|}~",
+      "alphabet": "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz!#$%&()*+:;<=>?@^_`{|}~",
       "keySize": 192,
       "msgSize": 0,
       "radix": 85,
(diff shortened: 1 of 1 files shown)


cpu:
I don't have a good idea of who to tag to review this given how niche AES FF1 is. In other circumstances I'd ask the OP of the issue, but in this case they wrote the tooling to generate the updated vectors and I've taken them as-is so it seems minimally helpful. 

While adopting the content I did of course perform my own review that the original issue was accurately described, and the fix as applied make sense to me and matches the description. Beyond that I think we shouldn't block this on further review and should instead merge and address any new issues that arise after the fact. Unless someone shares feedback that they'd prefer to see this have further review I'll plan to merge it ~tomorrow.

PR #242, opened 2026-06-02
ml-dsa: additional test vectors

tob-scott-a:
This PR adds new test cases to kill implementation bug classes identified in [DJB's latest loquacious screed against ML-DSA](https://cr.yp.to/papers/mldsa-20260601.pdf).

It's sad that Dr. Bernstein prioritizes trying to win cheap points in Internet debates by citing [CVEs against pre-production software](https://github.com/RustCrypto/signatures/security/advisories/GHSA-hcp2-x6j4-29j7) and fearmongering instead of **doing anything to actually help improve Internet security**.

To wit: If a CRQC were to exist today, Ed25519 would contribute zero security, and thus ML-DSA implementation security would be all that users can rely on.

So, I'm doing the work DJB should have done instead of writing that doc, using the [vector-forge skill](https://github.com/trailofbits/skills/tree/main/plugins/trailmark/skills/vector-forge).

-----

## Target

ML-DSA-44, ML-DSA-65, and ML-DSA-87 signing test vectors.

The target bug classes come from Daniel J. Bernstein, "Exploiting ML-DSA
bugs" (2026-06-01), <https://cr.yp.to/papers/mldsa-20260601.pdf>, and its
supplement, <https://cr.yp.to/2026/mldsabugs-20260601.tar.gz>.

## Bug Classes

- `SigningRandomnessExpansion`: signers that incorrectly expand the
  per-signature mask `y`, including AABBCC, ABABCDCD, and A0B0C0-style
  coefficient patterns.
- `RepeatedNonce`: signers that omit message-dependent material from nonce
  derivation, causing repeated `y` values across messages.
- `ClearedSecretK`: signers or key-generation paths that clear or ignore the
  secret `K` value used in per-signature mask derivation.

These are signing-side bugs: a faulty signer can emit signatures that verify
under a correct verifier. Verification-only tests do not catch them; exact
known-answer signing tests do.

## Vectors Added

Each parameter set gets three same-key deterministic signing KATs using
`rnd = 00...00`, added to both seed-key and full-key signing corpora.

| File | Added tcIds |
|---|---:|
| `testvectors_v1/mldsa_44_sign_seed_test.json` | 90-92 |
| `testvectors_v1/mldsa_44_sign_noseed_test.json` | 73-75 |
| `testvectors_v1/mldsa_65_sign_seed_test.json` | 109-111 |
| `testvectors_v1/mldsa_65_sign_noseed_test.json` | 78-80 |
| `testvectors_v1/mldsa_87_sign_seed_test.json` | 100-102 |
| `testvectors_v1/mldsa_87_sign_noseed_test.json` | 69-71 |

## Validation

- Generated deterministic signatures with a Python port of the paper
  supplement's `mldsa.sage` reference implementation.
- Reproduced the existing Wycheproof ML-DSA signing baseline with the same
  implementation and `rnd = 00...00` before generating new vectors.
- Verified all new signatures with the independent `pqcrypto` ML-DSA
  implementation.
- Validated the six changed JSON files against their declared schemas.

## Fault Replay

The new seed-key vectors were replayed against five simulated faulty signers: AABBCC, ABABCDCD, A0B0C0, truncated nonce derivation (`rho = H(K)`), and zeroed `K`.

| Parameter set | Fault models killed by exact KAT | Faulty signatures still verified |
|---|---:|---:|
| ML-DSA-44 | 5/5 | 5/5 |
| ML-DSA-65 | 5/5 | 5/5 |
| ML-DSA-87 | 5/5 | 5/5 |

This confirms the intended Vector Forge property: these vectors catch implementation bugs that ordinary sign-and-verify or interoperability tests would miss.


sgmenda (approved):
haven’t independently verified them, but lgtm.

sgmenda:
It is not required for merging, but @tob-scott-a could you share the precise sage script used to generate the test vectors?

testvectors_v1/mldsa_44_sign_noseed_test.json
```diff
@@ -31,9 +31,21 @@
     "Internal": {
       "bugType": "BASIC",
       "description": "A test vector with mu and no msg for use with Sign_internal and Verify_internal"
+    },
+    "SigningRandomnessExpansion": {
```
sgmenda, line 35: It might be worth citing the paper here too

tob-scott-a:
> It is not required for merging, but @tob-scott-a could you share the precise sage script used to generate the test vectors?

https://gist.github.com/tob-scott-a/2343b9e8e5980ecca25ae1d987f71a2b

testvectors_v1/mldsa_44_sign_noseed_test.json
tob-scott-a: Good catch. Done in https://github.com/C2SP/wycheproof/pull/242/commits/3f761d03f535b82d0482a409d6dafb12d84664e6

davidben (approved):
Verified by running these through our implementation.

FiloSottile:
I don't understand what these vectors add, I believe every single existing vector would catch Bernstein's made-up bugs.

https://mailarchive.ietf.org/arch/msg/tls/p5j6UCQGBAOblWPAjXb5ycMjxA4/

davidben:
Oh, if they're redundant then yeah I agree there's no point in adding them. (Haven't checked myself. I did not simulate the bugs to check if the existing vectors would have caught them or anything.)

tob-scott-a:
The implementations I worked with did not use the full Wycheproof test suite due to feature limitations (e.g., no deterministic support, only hedged), so the gaps I observed could have been a false negative. Codex agrees with your statement, @FiloSottile. Happy to close this if you think it adds no real value.

Issue #240, opened 2026-05-12
Tooling to merge and update test vectors


davidben:
Jumping off of some of the side discussion in https://github.com/C2SP/wycheproof/issues/239, and also what I've observed myself, some way to update and add test vectors would be handy.

What I've been doing is including my scripts in PRs like https://github.com/C2SP/wycheproof/pull/238. But those scripts have to import existing tests and modify them in-place, then keep all the IDs and test counts updated. This is a bit tedious. Maybe we can build a tool that improves this.

Thinking aloud, now that we have the `source` field, perhaps the tool could work something like:

    go run ./tools/vectormerge testvectors_v1/foo_test.json -import my_new_tests.json \
        -source github/davidben/my-fancy-new-tests -version 0.1

The tool would:

1. Check the schema of `my_new_tests.json` and error if anything is wrong, but maybe be permissive about things like `tcId`, `numberOfTests`, `source`, undefined `notes`, etc., since we'll fix them anyway.
2. Look for all existing tests whose source is `github/davidben/my-fancy-new-tests`. If the version is greater than 0.1, return an error. Otherwise, delete them.
3. Take all tests from `my_new_tests.json`. Insert them at...
    * The first position of the first deleted test, if any tests were removed
    * The end if no existing tests were removed
4. Set the `source` of any inserted tests with the specified source and version
5. Renumber `tcId` and fix `numberOfTests`
6. Merge the `notes` fields, warn or error if there are any conflicts
7. Format the resulting JSON canonically

Then a generator's job is to only output standalone JSON files, in whatever formatting is convenient, and the actual merge process is taken care of by a common script.

Thoughts?

cpu:
100% agree this would be a useful next step and preferable to trying to resurrect and maintain centralized generation code. I tried to hint at this in the [development priorities](https://github.com/C2SP/wycheproof#development-priorities) but haven't had time to work on it myself.

The workflow you're sketching out matches what I've had in my mind, though I've been a bit skeptical that Go's JSON capabilities would let us do this without bumping into the canonicalized field order/formatting nits from https://github.com/C2SP/wycheproof/issues/162 Maybe we need to push on consensus on whether a fixed ordering of fields would be acceptabe as a pre-req to merge tooling, or target Python as a first prototype.

davidben:
Yeah, the field order is a pain. I don't feel strongly. Using Python seems fine, as does just not caring and declaring that field order is now alphabetical.

cpu:
[...]

PR #243, opened 2026-06-02
Malformed PKCS8/SPKI coverage for ML-DSA-44/65/87

cpu:
Adds a schema + vectors that test a variety of malformed PKCS8/SPKI properties for ML-DSA. 

Most of the tests are not specific to ML-DSA, and indeed the schema is meant to be general enough for testing other algorithms similarly using the same vector generation methodology. I started with ML-DSA because some of these cases came up while discussing ML-DSA support in the Go standard library and we realized landing vectors here could be a more helpful contribution to the wider community. I wanted to share this work for early feedback before extending it too much further, but did attempt to write the generator code in a way that would facilitate that. Assuming I'm not way off track I will look to extend this for ECDSA, Ed25519 and RSA when time permits.

The current WIP generator code is in the [`cpu-pkcs8-pkix-suite-wip`](https://github.com/cpu/wycheproof/tree/cpu-pkcs8-pkix-suite-wip) branch of my fork (notably [`tools/pkixkeygen/`](https://github.com/cpu/wycheproof/tree/cpu-pkcs8-pkix-suite-wip/tools/pkixkeygen)). I will likely rewrite history there as I go, but will provide a stable ref here before merging these vectors.


cpu:
@davidben This might be up your alley and I'd appreciate any feedback you were able to share :bow:

cpu:
I'll reopen this once I have a chance to extend it to other algs, and to implement the generator with the bits from #240

PR #244, opened 2026-06-03
[ML-DSA] Add test vectors that exercise the c_tilde check in signature validation

gendx:
If the check is incorrectly implemented and doesn't compare the last byte of `c_tilde`, then these signatures are accepted even though they are invalid.

Note: contrary to existing test vectors labeled as `signature with a bit flip in c_tilde`, this new case really exercises only the `c_tilde` comparison at the end of the verification function. Indeed, this new synthetic case ensures that `w_approx` is zero regardless of the value of `c_tilde` (similarly to https://github.com/C2SP/wycheproof/pull/234). This allows fine-grained control on checking that each byte of `c_tilde` is taken into account in the comparison. Here I've only added one case where one bit in the last byte has been flipped (starting from the valid test labeled `signature that calls use_hint(1, 0)`), but one could similarly construct test cases that exercise each byte or bit of `c_tilde`.


cpu:
> Vector Lint / govulncheck (pull_request) - Failing after 1m

Thanks for the PR. I just wanted to note you can ignore this `govulncheck` finding. I'll fix the flagged deps in the tooling separately and you can rebase it away afterwards.

Edit: see https://github.com/C2SP/wycheproof/pull/245

cpu (approved):
Thanks! I verified that introducing the buggy `c_tilde` check wasn't caught by the existing `mldsa_*_verify_test.json‎
` vectors, but _was_ caught by your additions. 👍 

@gendx Would you mind rebasing on `main` so CI pases?

>  Here I've only added one case where one bit in the last byte has been flipped (starting from the valid test labeled signature that calls use_hint(1, 0)), but one could similarly construct test cases that exercise each byte or bit of c_tilde.

That sounds interesting to me. Would you be up for submitting vectors across the byte range of `c_tilde` or do you think it's excessive?

FiloSottile:
Great catch! Definitely worth having at least a vector for the first byte, if not all of them.

cpu:
I'm going to merge this now that it's rebased, but would welcome another PR with vectors for the first byte as well. 

Thanks again!

gendx:
> I'm going to merge this now that it's rebased, but would welcome another PR with vectors for the first byte as well.

I was already working on them, opened https://github.com/C2SP/wycheproof/pull/246 with all the bytes now :)

PR #246, opened 2026-06-04
[ML-DSA] Add test vectors for each byte of c_tilde.

gendx:
Follow-up from https://github.com/C2SP/wycheproof/pull/244.


cpu (approved):
Great work, thank you.

I ginned up a quick patch to the Go ML-DSA impl that let me specify which byte to skip comparing in `c_tilde` and marched across the whole range for ML-DSA-44/-65/-87 and verified the expected vector failed in each case (32 distinct failures for -44, 48 for -65 and 64 for -87). I also smoke tested that the vectors all pass with the original non-buggy logic restored.

PR #247, opened 2026-06-05
Adding bouncy castle rust to listed downstream projects

ounsworth:
Two implementations each of ML-DSA and ML-KEM now have wycheproof test harnesses and are fully passing:
* [ML-DSA](https://github.com/bcgit/bc-rust/tree/release/0.1.2alpha/crypto/mldsa/tests)
* [ML-DSA-lowmemory](https://github.com/bcgit/bc-rust/tree/release/0.1.2alpha/crypto/mldsa_lowmemory/tests)
* [ML-KEM](https://github.com/bcgit/bc-rust/tree/release/0.1.2alpha/crypto/mlkem/tests)
* [ML-KEM-lowmemory](https://github.com/bcgit/bc-rust/tree/release/0.1.2alpha/crypto/mlkem_lowmemory/tests)

Additionally, this PR adds one test case that I constructed to make my mutation tests happy, and I believe is not currently in the wycheproof tests: a test case for ML-DSA-44 containing a signature where coeffs are all in range (1<<17 -1), but the norm of z exceeds GAMMA1 - BETA.

Finally, I have a question about some of the test vectors in mlkem_512_encaps_test.json:
The set of "ModulusOverflow" tests provide an m that is 64 bytes, but FIPS 203 Alg 14 says that m is 32 bytes. So when I run this, my implementation rejects because it's the wrong length of m. Is that really what that test is supposed to be testing? If yes, do we really need 9 different tests all testing that 64 != 32?


doc/bugs.md
```diff
@@ -85,3 +85,9 @@
 
 [ToB elliptic]: https://blog.trailofbits.com/2025/11/18/we-found-cryptography-bugs-in-the-elliptic-library-using-wycheproof/
 [elliptic#317]: https://github.com/indutny/elliptic/pull/317
+
+## Package Bouncy Castle Rust
```
cpu, line 89: In the C2SP adoption updates I re-titled this page "Historic" bugs with the anticipation we aren't interested in maintaining a trophy case going forward. I'm happy to hear the vectors were useful for your project but would personally vote to drop this part of the diff unless another contributor feels strongly.

testvectors_v1/mldsa_44_verify_test.json
```diff
@@ -1850,6 +1850,17 @@
             "BoundaryCondition",
             "ZeroPublicKey"
           ]
+        },
+        {
+          "tcId": 150,
```
cpu, line 1855: Adding this testcase also requires updating the `numberOfTests` on L53. That should fix the [vectorlint failure in CI](https://github.com/C2SP/wycheproof/actions/runs/26996100004/job/79666066484?pr=247).

cpu (commented):
Thanks for the PR.

Without having looked closely at the new vector addition, or the MLKEM question, here's some initial feedback.

Can you also split this into discrete commits? I think ideally it would be one commit for the `README.md` update and a second one with a descriptive message for the `mldsa_44_verify_test.json` addition.

ounsworth:
@cpu Thanks for the feedback.

I will split the readme changes and the new test vector into separate PRs, remove the trophy case, put the MLKEM question into an Issue, and and I'll run the linter. 
Looking at my own tests, I have a few valid tests that also exercise the GAMMA1 - BETA constant, so I'll add those too.

PR #253, opened 2026-06-06
Mldsa z norm test

ounsworth:
This PR adds one test case that I constructed to make my mutation tests happy, and I believe is not currently in the wycheproof tests: a test case for ML-DSA-44 containing a signature where coeffs are all in range (1<<17 -1), but the norm of z exceeds GAMMA1 - BETA (FIPS 204 Alg 8 line 13)


testvectors_v1/mldsa_44_verify_test.json
```diff
@@ -2224,6 +2224,28 @@
...
+          "result": "invalid",
+          "flags": [
```
cpu, line 2243: I think this vector should also include `InfinityNormViolation` no?

testvectors_v1/mldsa_44_verify_test.json
```diff
@@ -2224,6 +2224,28 @@
...
+      "source": {
+        "name": "github/ounsworth",
```
cpu, line 2233: I think it'd be helpful to qualify this further so that in the future if you contributed other vectors it would be unambiguous. 

E.g. sophie used `github/sophieschmieg/infinitynorm`

cpu (commented):
Thanks.

On the administrative side, your PR title and branch name mention ML-KEM, but this is a ML-DSA vector right? I think we can ignore the branch name, but the PR title should be updated. It also looks like you included the README update commit from https://github.com/C2SP/wycheproof/pull/252 Could you rebase on `main` so that falls away?

On the vector side: can you share the code you used to generate this vector? It's helpful when reviewing the vector contribution (but doesn't need to be fancy code optimized for maintenance or anything like that).

> a test case for ML-DSA-44 containing a signature where coeffs are all in range (1<<17 -1), but the norm of z exceeds GAMMA1 - BETA (FIPS 204 Alg 8 line 13)

ISTM that the new vector does have an infinity norm out of range as claimed, but it's not separating out the two conditions from FIPS-204 Alg 8 L13. An implementation with a buggy/missing norm check will still reject this signature on the c-hash check alone.

When I ran this through my own implementation c-tilde from the signature is `cb8f8b46d73e7c273500555acbe0e7cf1da54d950675248e11bff5940b45f52f` but c-prime is `20a8df0e43e8bc76b6eb94c47dde80184d8230b53503cfe5937dcf483a1d4e0f` so the signature is rejected regardless of whether the norm check runs.

To isolate the norm check, the vector would need to be constructed so that c̃ == c′ but ‖z‖∞ ≥ γ1 − β

I'm also not sure this is a condition that's specific to ML-DSA-44. Wouldn't we need similar coverage for `mldsa_65_verify_test.json` and `mldsa_87_verify_test.json` or is your assertion that we have coverage of this case there but missed it for -44? It'd be helpful to determine that conclusively before adding a vector only for -44.

FiloSottile:
I am confused, isn't this what tcId 125 and 126 target?

ounsworth:
> I am confused, isn't this what tcId 125 and 126 target?

Weird. I swear I CTRL+F'd for "gamma" and didn't find anything.

You're right, this is ML-DSA, not ML-KEM. I'm clearly moving too fast and not paying close enough attention.

Sorry for wasting your time. I'll close this PR.

ounsworth:
The code that generated this vector is that I printed out a valid signature in hex, changed some of the bytes to larger values, and then put it back it. Which, yes, is a broken signature because c-prime != c-tilde, but in my code, it makes my mutation tests happy because I check the norm of the signature before computing c-prime and c-tilde. So in my test suite, this test is doing something useful.

I only did it for mldsa-44 because my -44 / -65 / -87 are the same implementation and one test was enough to pin my mutant; but I could easily construct an example for the other parameter sets if you want.

I see your point that this would be a better test if c̃ == c′ but ‖z‖∞ ≥ γ1 − β, but I have no idea how to construct that since c_prime comes from SHAKE256(w1-prime), and w1-prime comes from the whole matrix operations which at the very top depend on c-tilde, so you would need to construct the busted signature value such that all the public key operations gives the identical w1-prime. So I think that constructing the test case you're asking for requires first finding a full mathematical break for the ML-DSA algorithm? But you're right that this is going to fail the signature check regardless because the C's don't match.

I somehow missed that you already have the gamma1 norm check covered in tcid 125 & 126, so I'm happy to just drop this PR. Sorry for the wasted time.

PR #254, opened 2026-06-10
Add ML-KEM re-encryption test vectors.

lukaszobernig:
This adds ML-KEM test vectors that fail to decapsulate on a correct implementation but decapsulate to the correct shared secret on a faulty implementation. The first vector tests whether the implementation does not compare `c_1` and `c_1'` correctly, specifically whether the implementation forgets to compare the last byte. The second vector tests the same for the comparison of `c_2` and `c_2'`.


cpu:
Thanks for the PR! 

While reviewing this & fiddling with introducing the c_1/c_2 comparison bugs in an implementation I realized that the original `mlkem_semi_expanded_decaps_test_schema.json` should have the expected `K` like the seed-form `mlkem_test_schema.json` already had. I think this was an oversight when we landed the semi-expanded schema/vectors initially. I went ahead and tacked on a commit that updates the schema + backfills the values (backfill tool [here](https://github.com/cpu/wycheproof/blob/2e522af1f7645e07da42a1212ab883f3f090c932/tools/mlkem_semi_expanded_backfill/main.go)).

I also had some review feedback for your new vectors beyond adding `K` and since I was pushing commits I applied it directly. Please push-back if you disagree with the changes.

1. I think these should be `result = valid` - the invalid cases so far have been instances where `Decapsulate` fails, and that isn't applicable here.
2. I think these vectors should be in a new group so that we can have a `source` distinct from the original `"github/aws/aws-lc"`. I went with `"github/lukaszobernig/reenc"` but if you'd prefer something else we can change it.

With all of the above handled I was able to verify that introducing a representative c_1/c_2 comparison bug in the Go ML-KEM implementation went unnoticed by existing vectors, but was caught by the expected new vectors :+1:

@sgmenda @FiloSottile Would appreciate you folks reviewing before I merge since I've made adjustments of my own. :bow:

cpu (approved):

lukaszobernig:
Thanks for making those changes, I wasn't super sure how to include the vectors into the existing schema. But what you're proposing sounds good to me.

sgmenda (approved):
the changes including @cpu 's improvements look great to me.

also ran them against aws-lc with no issues.

thanks so much for the contribution.

schemas/mlkem_semi_expanded_decaps_test_schema.json
```diff
@@ -78,11 +78,21 @@
...
               "c": {
                 "type": "string",
                 "format": "HexBytes",
                 "description": "An input ciphertext"
               },
+              "K": {
+                "type": "string",
+                "format": "HexBytes",
+                "description": "If present, the shared key the implementation MUST return on a successful Decapsulate call."
+              },
```
FiloSottile, line 95: Make this required for result: valid?

FiloSottile (approved):

schemas/mlkem_semi_expanded_decaps_test_schema.json
cpu: Done.

As a side-note, adding these sort of conditional schema rules seems to interact badly with some code generators. As a concrete example the one I've been using for JSON Schema -> Go will stop emitting a typed structure for anything that has a conditional rule, instead spitting out an `interface{}` that we have to cast to a manually defined structure.

I'm operating on the assumption we care more about vector accuracy vs fidelity of code generation so adding this kind of conditional validation is the right call and we'll deal with the less useful code generation downstream, but I wanted to mention it as an explicit trade-off. _(It's also possible other JSON schema tools handle it better, I haven't done an exhaustive survey)_.

FiloSottile: Ah, that's unfortunate, but yeah not having to manually notice missing fields feels more important.

This sounds very fixable, btw. Want me to point @filippo-claude at it?

cpu: > This sounds very fixable, btw. Want me to point @filippo-claude at it?

Sure! If you want to take a crack at it I've been using https://github.com/omissis/go-jsonschema Maybe there's an easy fix.

<details>
<summary>tools/schemagen.go in this repo is already setup and just needs a small diff to output the generated code instead of just checking that it builds:</summary>

```diff
diff --git a/tools/schemagen/schemagen.go b/tools/schemagen/schemagen.go
index 5b656c8..3c8fd4f 100644
--- a/tools/schemagen/schemagen.go
+++ b/tools/schemagen/schemagen.go
@@ -17,7 +17,10 @@ import (
        "github.com/atombender/go-jsonschema/pkg/generator"
 )

-var schemaDirectory = flag.String("schemas-dir", "schemas", "directory containing schema files")
+var (
+       schemaDirectory = flag.String("schemas-dir", "schemas", "directory containing schema files")
+       dumpStdout      = flag.Bool("stdout", false, "write the generated source to stdout")
+)

 func main() {
        flag.Parse()
@@ -61,11 +64,17 @@ func main() {
        if sourceCount := len(sources); sourceCount != 1 {
                log.Fatalf("expected to generate 1 source file, got %d\n", sourceCount)
        }
-       _, ok := sources[ouputName]
+       source, ok := sources[ouputName]
        if !ok {
                log.Fatalf("missing generated %q output file source", ouputName)
        }

+       if *dumpStdout {
+               if _, err := os.Stdout.Write(source); err != nil {
+                       log.Fatalf("writing source to stdout: %v", err)
+               }
+       }
+
        for _, warning := range warnings {
                log.Printf("⚠ Warning: %s", warning)
        }
```

</details>

If you apply that you can get a nice loop going by deleting [schemas/mlkem_semi_expanded_decaps_test_schema.json#L116-L129](https://github.com/lukaszobernig/wycheproof/blob/c2bd219a724cf40d49e26135f049cbf6ef54c2b0/schemas/mlkem_semi_expanded_decaps_test_schema.json#L116-L129), running `tools/schemagen.go --stdout > before.go`, then reverting the schema back and running it again. You'll see the nice structure drop-out of the generated code and get replaced with `interface{}`.

cpu:
Thanks again @lukaszobernig!

schemas/mlkem_semi_expanded_decaps_test_schema.json
FiloSottile: https://github.com/omissis/go-jsonschema/pull/590

Issue #262, opened 2026-07-02
Generate SP 800-232 compatible Ascon-AEAD128 test vectors


smuellerDD:
As reported and validated in [1], the Ascon test vectors apply to the NIST finalist proposal of Ascon, but not SP800-232, i.e. the standardized version of AsconAEAD-128.

May I propose to include the updated Ascon vector set from [1]?

[1] https://github.com/bleichenbacher-daniel/Rooterberg/issues/8

cpu:
Thanks for reporting. I think short-term we could certainly clarify that the vectors in-repo are for a pre-standardization version to help avoid confusion. I also agree with you that having vectors for the standardized version would be an excellent contribution. Are you interested in making a PR?

> May I propose to include the updated Ascon vector set from [1]?

As mentioned on the other issue, I don't believe that would be workable in the current state as the vectors in [1] are not under a compatible license. If they were available under an Apache-2.0 compatible license I would be happy to review a PR from you, or the original author that imports them. Ideally matching the general shape of the pre-existing vectors unless there was a strong reason it wasn't workable.

cpu:
> the Ascon test vectors apply to the NIST finalist proposal of Ascon, but not SP800-232

Just to verify: this applies to all of `testvectors_v1/ascon80pq_test.json`,  `testvectors_v1/ascon128_test.json`, and `testvectors_v1/ascon128a_test.json`?

cpu:
> Just to verify: this applies to all of testvectors_v1/ascon80pq_test.json, testvectors_v1/ascon128_test.json, and testvectors_v1/ascon128a_test.json?

Answering my own question: yes, and AFAICT SP 800-232 only specified Ascon-AEAD128 based on the v1.2 Ascon128a. I don't think there's  an equivalent to the 64-bit rate or 160-bit key variants. If I'm mistaken, constructive feedback on https://github.com/C2SP/wycheproof/pull/263 is welcome. I have no direct experience with these lightweight cryptography algorithms.

I think it won't be especially difficult to generate similar SP 800-232 vectors, and I propose we do that under an unambiguous filename like `testvectors_v1/ascon_sp800_232_aead128_test.json` using the pre-existing schema `aead_test_schema_v1.json`. I don't know when I'll have time to tackle that personally (again, PRs welcome!), but in the meantime #263 can hopefully avoid further confusion. Once it lands I will re-title this issue as a request for SP 800-232 vectors

smuellerDD:
[...]

PR #263, opened 2026-07-03
ascon: clarfiy vectors are for v1.2 (not SP 800-232)

cpu:
Add mention to the `ascon128_test.json`, `ascon128a_test.json`, and `ascon80pq_test.json` vector headers that they implement Ascon-128, Ascon-128a, and Ascon-80pq v1.2, the final-round submission to the NIST LWC competition. They are **not** interoperable with the SP 800-232 standardized algorithm.

For reference, see [ascon.py](https://github.com/C2SP/wycheproof/blob/cff6adf42662469a1871e57303a0ad1d758ed8c0/src/ascon.py) and [gen_ascon.py](https://github.com/C2SP/wycheproof/blob/cff6adf42662469a1871e57303a0ad1d758ed8c0/src/gen_ascon.py) from [google-wycheproof/v0.9](https://github.com/C2SP/wycheproof/releases/tag/google-wycheproof%2Fv0.9).

We should replace/augment these with SP 800-232 compatible vectors as follow-up. In the meantime, try to avoid potential confusion.

Updates https://github.com/C2SP/wycheproof/issues/262

### DIFF
--- a/testvectors_v1/ascon128a_test.json
+++ b/testvectors_v1/ascon128a_test.json
@@ -3,6 +3,12 @@
   "schema": "aead_test_schema_v1.json",
   "numberOfTests": 192,
   "header": [
+    "These test vectors implement Ascon-128a v1.2, the final-round submission to",
+    "the NIST Lightweight Cryptography competition. Ascon-AEAD128 as standardized",
+    "in NIST SP 800-232 is derived from Ascon-128a but is NOT interoperable with",
+    "it: initialization vector, byte ordering, padding, and domain separation are",
+    "different. As a result these vectors are not suitable for testing an SP 800-232",
+    "implementation.",
     "Test vectors of type AeadTest test authenticated encryption with additional data.",
     "The test vectors are intended for testing both encryption and decryption.",
     "Test vectors with \"result\" : \"valid\" are valid encryptions.",
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ including
 - AES-GCM
 - AES-SIV
 - ARIA
-- ASCON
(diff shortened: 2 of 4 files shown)


cpu:
@C2SP/wycheproof-reviewers I think this is a pretty easy review if someone has time.

testvectors_v1/ascon128a_test.json
```diff
@@ -3,6 +3,12 @@
   "schema": "aead_test_schema_v1.json",
   "numberOfTests": 192,
   "header": [
+    "These test vectors implement Ascon-128a v1.2, the final-round submission to",
+    "the NIST Lightweight Cryptography competition. Ascon-AEAD128 as standardized",
+    "in NIST SP 800-232 is derived from Ascon-128a but is NOT interoperable with",
+    "it: it uses a different initialization vector, byte ordering, padding, and",
+    "domain separation. These vectors are not suitable for testing an SP 800-232",
```
botovq: Not a native speaker, but "it uses a different" followed by a list seems a bit odd. Perhaps turn it around, also avoiding the doubled "it":

```suggestion
    "it: initialization vector, byte ordering, padding, and domain separation",
    "are different. These vectors are not suitable for testing an SP 800-232",
```

botovq (approved):
Seems fine. One small suggestion if that makes sense to you.

testvectors_v1/ascon128a_test.json
cpu: Good point, I like your suggested wording :+1:  Updated.

botovq (approved):

PR #265, opened 2026-07-13
Add c2sp.org/chunked-encryption (Cobblestone) test vectors

FiloSottile:
/cc @alex @reaperhulk


alex:
Perfect, thanks! Our current PR uses the ones from your impl.

FiloSottile:
These are ported from those, and the Go implementation (now renamed to filippo.io/cobblestone) now uses these.

schemas/c2sp_chunked_encryption_schema.json
```diff
@@ -0,0 +1,302 @@
...
+          "items": {
+            "$ref": "#/definitions/ChunkedEncryptionTestVector"
+          }
+        },
+        "source": {
+          "$ref": "common.json#/definitions/Source"
+        }
+      },
+      "additionalProperties": false,
+      "required": [
```
cpu, line 36: I think `source` should also be a required field.

schemas/c2sp_chunked_encryption_schema.json
```diff
@@ -0,0 +1,302 @@
...
+        "aead": {
+          "enum": [
+            "AEAD_AES_128_GCM",
+            "AEAD_AES_256_GCM"
+          ],
+          "description": "The underlying AEAD, by IANA registry name"
+        },
+        "sha": {
+          "type": "string",
+          "format": "MdName",
```
cpu: There's no validator registered for `MdName` in `vectorlint` so this is fairly decorative.

I think unless you plan to convert other schemas to use a defined+linted `MdName` format maybe this should just be an enum with some variants listed? Or should it be fixed to "SHA-512" since there are no chunked vectors using a different digest at this point.

schemas/c2sp_chunked_encryption_schema.json
```diff
@@ -0,0 +1,302 @@
...
+        "ctx",
+        "ct",
+        "result",
+        "flags"
+      ]
+    }
+  },
+  "properties": {
+    "algorithm": {
+      "type": "string",
```
cpu: Can this be an enum with fixed values, something like: `"enum": ["Cobblestone-128", "Cobblestone-256"] `

schemas/c2sp_chunked_encryption_schema.json
```diff
@@ -0,0 +1,302 @@
...
+          "format": "HexBytes",
+          "description": "The derived AEAD key, for testing raw mode directly (absent for invalid vectors that exercise key derivation or the commitment check)"
+        },
+        "baseNonce": {
+          "type": "string",
+          "format": "HexBytes",
+          "description": "The derived base nonce, for testing raw mode directly (absent for invalid vectors that exercise key derivation or the commitment check)"
+        },
+        "msgLength": {
+          "type": "integer",
```
cpu, line 81: worth adding a `"minimum": 0` here?

cpu (commented):
Here's some initial feedback mainly focused on the new schema.

reaperhulk:
I'm very keen on getting this merged so we can get our own cobblestone release out, anything I can do to help Filippo?

FiloSottile:
I agree with all of @cpu's comments and just haven't gotten to it, so if you have the cycles, feel free to make a PR to the push-nkskznqqunmy branch addressing them and I or @cpu will turbomerge it :)

cpu:
I'm away from a full keyboard atm but I can apply the feedback tomorrow morning if nobody beats me to it.

schemas/c2sp_chunked_encryption_schema.json
cpu: Fixed.

schemas/c2sp_chunked_encryption_schema.json
cpu: Switched to an enum with just "SHA-512" listed.

schemas/c2sp_chunked_encryption_schema.json
cpu: Fixed.

schemas/c2sp_chunked_encryption_schema.json
cpu: Added.

cpu:
@reaperhulk All set!

reaperhulk:
Thanks! Beat me to it 😄 very much appreciated

PR #269, opened 2026-08-11
Add missing schemas, source attribution

cpu:
This PR adds JSON schemas that were referenced from `testvectors_v1/` but didn't exist. These were an artifact from the original Google version of the project that we've been ignoring in the vector lint tooling. Having complete schemas is helpful for catching errors when we extend the vectors, and also allows consumers to do code generation for loading the vector data.

I've made a best effort to match up what exists in the test vectors, the original Python generator code from [the tag w/ generator source](https://github.com/C2SP/wycheproof/releases/tag/google-wycheproof%2Fv0.9), and new schema content. There's likely room for further improvement down the road but for now I've limited myself to a first pass.

This will conflict with https://github.com/C2SP/wycheproof/pull/268 (and I had to cherry-pick the line-ending fix in here). I'll resolve in whichever direction makes sense based on merge order.


cpu:
@C2SP/wycheproof-reviewers would love a review here as well if someone can spare the time :-)

rolandshoemaker (approved):
This... seems right to me?

PR #271, opened 2026-08-19
ed25519, ristretto: more test coverage

tob-scott-a:
Adds 146 regression vectors derived from cryptographic boundary fixes in `sodium_compat` [v2.5.1](https://github.com/paragonie/sodium_compat/releases/tag/v2.5.1) and [v2.5.2](https://github.com/paragonie/sodium_compat/releases/tag/v2.5.2), using [vector-forge](https://github.com/trailofbits/skills/tree/main/plugins/trailmark/skills/vector-forge):

- 59 Ed25519 small- and mixed-order subgroup cases
- 34 Ed25519-to-X25519 conversion cases
- 42 ChaCha20/XChaCha20 counter carry and overflow cases
- 4 Ristretto255 scalar-inversion cases
- 7 Secretstream counter and rekey state checkpoints

This also adds four schemas for the new vector formats.

The main idea behind this PR was to write stable test vectors to exercise the attack surface in those releases so that other implementations that currently test with Wycheproof can ensure they are resilient to similar bugs.

### DIFF
--- a/schemas/eddsa_to_xdh_schema_v1.json
+++ b/schemas/eddsa_to_xdh_schema_v1.json
@@ -1,0 +1,109 @@
+{
+  "type": "object",
+  "definitions": {
+    "EddsaToXdhTestGroup": {
+      "type": "object",
+      "properties": {
+        "type": {
+          "enum": [
+            "EddsaToXdh"
+          ]
+        },
+        "source": {
+          "$ref": "common.json#/definitions/Source"
+        },
+        "edwardsCurve": {
+          "type": "string",
+          "format": "EcCurve",
+          "description": "the source Edwards curve"
+        },
+        "montgomeryCurve": {
+          "type": "string",
+          "format": "EcCurve",
+          "description": "the destination Montgomery curve"
+        },
+        "tests": {
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/EddsaToXdhTestVector"
+          }
+        }
+      },
+      "required": ["type", "source", "edwardsCurve", "montgomeryCurve", "tests"],
+      "additionalProperties": false
+    },
+    "EddsaToXdhTestVector": {
+      "type": "object",
+      "properties": {
+        "tcId": {
+          "type": "integer",
+          "description": "Identifier of the test case"
+        },
+        "comment": {
+          "type": "string",
+          "description": "A brief description of the test case"
+        },
+        "publicKey": {
+          "type": "string",
+          "format": "HexBytes",
+          "description": "the encoded Edwards public key"
+        },
+        "convertedPublicKey": {
+          "type": "string",
+          "format": "HexBytes",
+          "description": "the expected encoded Montgomery public key"
+        },
+        "result": {
+          "$ref": "common.json#/definitions/Result"
+        },
+        "flags": {
+          "type": "array",
+          "items": {
+            "type": "string"
+          },
+          "description": "A list of flags"
+        }
+      },
+      "required": ["tcId", "comment", "publicKey", "result", "flags"],
+      "additionalProperties": false
+    }
+  },
+  "properties": {
+    "algorithm": {
+      "type": "string",
+      "description": "the primitive tested in the test file"
+    },
+    "generatorVersion": {
+      "type": "string",
+      "description": "DEPRECATED: prefer \"source\" property in test group",
+      "deprecated": true
+    },
+    "header": {
+      "type": "array",
+      "items": {
+        "type": "string"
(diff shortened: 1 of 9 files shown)


cpu:
Thank you! This is in my list to review when I'm back from a short vacation, later next week.

schemas/eddsa_to_xdh_schema_v1.json
```diff
@@ -0,0 +1,109 @@
...
+  "properties": {
+    "algorithm": {
+      "type": "string",
+      "description": "the primitive tested in the test file"
+    },
+    "generatorVersion": {
+      "type": "string",
+      "description": "DEPRECATED: prefer \"source\" property in test group",
+      "deprecated": true
+    },
```
cpu, line 80: No need to add this for new schemas (same feedback for the other new schema files).

schemas/ristretto_scalar_test_schema_v1.json
```diff
@@ -0,0 +1,109 @@
...
+          "enum": [
+            "RistrettoScalarTest"
+          ]
+        },
+        "source": {
+          "$ref": "common.json#/definitions/Source"
+        },
+        "curve": {
+          "enum": [
+            "ristretto255"
```
cpu, line 17: might be nicer to avoid single-value enums here (and for operation).

schemas/eddsa_to_xdh_schema_v1.json
```diff
@@ -0,0 +1,109 @@
+{
```
cpu, line 1: You can drop the `_v1` portion of the file name for these new schemas. The only reason some of the files have that today is that there used to be a `testvectors/` schema and a `testvectors_v1/` schema for some types, but we've removed all the `testvectors/` stuff.

schemas/ristretto_scalar_test_schema_v1.json
```diff
@@ -0,0 +1,109 @@
...
+        },
+        "scalar": {
+          "type": "string",
+          "format": "HexBytes",
+          "description": "the scalar in canonical little-endian form"
+        },
+        "resultScalar": {
+          "type": "string",
+          "format": "HexBytes",
+          "description": "the expected inverse; omitted for invalid tests"
```
cpu, line 54: we can encode this "required except for invalid" notion into the schema itself. I think [the chunked encryption schema](https://github.com/C2SP/wycheproof/blob/dac1dd4729fd1f8dd9e1e9f3dce51d783da6c166/schemas/c2sp_chunked_encryption_schema.json) has an example.

Same feedback for the other instances.

testvectors_v1/ed25519_test.json
```diff
@@ -3355,6 +3368,1157 @@
...
+        "type": "EDDSAPublicKey",
+        "curve": "edwards25519",
+        "keySize": 255,
+        "pk": "0100000000000000000000000000000000000000000000000000000000000000"
+      },
+      "publicKeyDer": "302a300506032b65700321000100000000000000000000000000000000000000000000000000000000000000",
+      "publicKeyPem": "-----BEGIN PUBLIC KEY-----\nMCowBQYDK2VwAyEAAQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=\n-----END PUBLIC KEY-----\n",
+      "tests": [
+        {
+          "tcId": 152,
```
cpu, line 3388: I might be misunderstanding but I think tcIds 152 .. 180 would be better as `result: acceptable` with a flag instead of `result: invalid`. AFAIU RFC 8032 verification satisfies these vectors and this is encoding a stricter subgroup policy. 

As a concrete example, upstream libsodium 1.0.18 accepts 160..180, neither that version or current master performs a full subgroup validation.

WDYT? Am I missing something?

schemas/secretstream_test_schema_v1.json
```diff
@@ -0,0 +1,143 @@
...
+        },
+        "counter": {
+          "type": "string",
+          "format": "HexBytes",
+          "description": "the 32-bit counter in little-endian order"
+        },
+        "inonce": {
+          "type": "string",
+          "format": "HexBytes",
+          "description": "the 12-byte logical internal nonce"
```
cpu, line 20: I'm not sure I follow why this is 12 bytes instead of 8. It looks like all of the values have a four trailing zero bytes. I thought the nonce construction was `counter[4] || inonce[8]`.

cpu (commented):
Thanks! Here's a bit of feedback.

It might also be nice to update `README.md` with the new coverage.

PR #273, opened 2026-08-20
Add HPKE test vectors

sgmenda:
Wycheproof test vectors for HPKE (RFC 9180) base mode, covering DHKEM(X25519, HKDF-SHA256) and ML-KEM-512/768/1024 (draft-ietf-hpke-pq). See doc/hpke.md for detailed documentation.

- we mark X25519 all-zero-DH keys as *invalid* per RFC 9180 §5.1.1, which is stricter than RFC 7748 and than `x25519_test.json`, which marks them *acceptable*.
- the sequence files exercise HPKE's stateful, ordered Open, including out-of-order, replayed, and skipped ciphertexts.
- we generated these vectors using aws-lc and checked them against pyca/cryptography, Go crypto/hpke, and Bouncy Castle everywhere they overlap. notably, only aws-lc and Bouncy Castle implement ML-KEM-512 in HPKE.


cpu:
Thank you! Adding HPKE coverage is something I've talked about with Filippo but neither of us had gotten to it yet. I won't be able to review this until the ~end of next week but I'm very supportive.

As a side note I think the upstream hpkewg is interested in finding a home for their vectors (https://github.com/hpkewg/hpke/issues/72). Did you want to reach out and see if they're interested in joining forces and using c2sp/wycheproof for that home? They also recently added some new corner cases (https://github.com/hpkewg/hpke/pull/70), and if you haven't already considered those they might be helpful additions.

sgmenda:
@cpu thanks for the pointers, I will reach out.

My initial instinct was those vectors should go in CCTV. But from the perspective of a crypto library, it's easier to track one external test vector library than two (Wycheproof and CCTV). @FiloSottile I'm curious for your thoughts?

doc/hpke.md
```diff
@@ -0,0 +1,121 @@
...
+  over-strict.
+- **DHKEM(X25519)**: HPKE is stricter than plain X25519 here. A small-order
+  `pkRm` whose Diffie-Hellman output is all-zero MUST be rejected, because
+  RFC 9180 Section 5.1.1 requires `DH` to abort on an all-zero shared secret —
+  a check RFC 7748 leaves optional (`x25519_test.json` labels such keys
+  `acceptable`). Wrong-length keys are rejected too.
+
+The ML-KEM encapsulation files are a single-implementation canary: among the
+implementations used to cross-validate these vectors, only AWS-LC implements
+ML-KEM in HPKE.
```
jakemas, line 51: Did you mean to say only 512 in:
"only AWS-LC implements ML-KEM in HPKE." Because BoringSSL implements ML-KEM in HPKE, right?
```suggestion
implementations used to cross-validate these vectors, only AWS-LC implements
ML-KEM-512 in HPKE.
```

doc/hpke.md
```diff
@@ -0,0 +1,121 @@
...
+
+`skRm` is the serialized recipient private key (`SerializePrivateKey`): the
+64-byte `d || z` seed for the ML-KEM KEMs (`Nsk = 64`, the expanded key
+re-derived via `ML-KEM.KeyGen_internal`), and the 32-byte scalar for
+DHKEM(X25519) (`Nsk = 32`). `pkRm` is the serialized public (encapsulation)
+key; it can also be derived from `skRm`.
+
+Valid decryption cases also carry `ikmE`, the 32-byte seed that fixes the
+sender's ephemeral key: for ML-KEM the encapsulation randomness, for
+DHKEM(X25519) the ephemeral private key `skEm` (what a deterministic-
```
jakemas, line 91: This renders all on one line like "what a deterministic- encapsulation hook consumes, not RFC 9180's" so we don't need the `-`:


```suggestion
DHKEM(X25519) the ephemeral private key `skEm` (what a deterministic
```

schemas/hpke_encap_test_schema.json
```diff
@@ -0,0 +1,179 @@
...
+              },
+              "ikmE": {
+                "type": "string",
+                "format": "HexBytes",
+                "description": "Optional. The sender's encapsulation randomness (32 bytes for ML-KEM). Present for valid cases, where deterministic encapsulation reproduces enc and ct. Absent for invalid cases."
+              },
+              "skRm": {
+                "type": "string",
+                "format": "HexBytes",
+                "description": "Optional. The recipient private key (64-byte d || z seed) corresponding to pkRm, when one exists. Present lets a consumer additionally decapsulate enc and open ct to recover pt."
```
jakemas, line 112: Doesn't this schema apply to both ML-KEM and x25519? i.e. `hpke_x25519_encap_test.json` has a 32-byte skRm. In decap you cover both cases

cpu:
> My initial instinct was those vectors should go in CCTV. But from the perspective of a crypto library, it's easier to track one external test vector library than two (Wycheproof and CCTV)

Yeah I think the boundary of which belongs where is a little fuzzy. Beyond the benefit of one-source-of-truth Wycheproof 's JSON schemas make life easier for downstreams :man_shrugging: I'd be happy with either.

doc/hpke.md
jakemas: I see BC has 512 (and X-wing!) https://github.com/bcgit/bc-java/blob/main/core/src/main/java/org/bouncycastle/crypto/hpke/MLKEM.java

sgmenda:
I have pushed a major revision.

@jakemas thanks for the suggestions. now, every vector in this PR has been checked by at least two implementations.

@cpu thanks for the pointer to the hpkewg edge-case vectors. now, this PR includes two new classes of vectors:
1. vectors which test that the stateful `Open` correctly accepts valid sequences and rejects invalid ones.
2. vectors which test that zero-bytes in info/aad/pt are handled correctly

I have not imported the hpkewg vectors, since they would have expanded the scope significantly to include P-256, PSK, and exporting. Instead, I took inspiration from their edge case vectors to ensure that our coverage is a superset (for the schemes in scope).

schemas/hpke_encap_test_schema.json
```diff
@@ -0,0 +1,179 @@
...
+        "kem": {
+          "enum": [
+            "ML-KEM-512",
+            "ML-KEM-768",
+            "ML-KEM-1024",
+            "DHKEM(X25519, HKDF-SHA256)"
+          ]
+        },
+        "kemId": {
+          "type": "integer"
```
cpu, line 64: minor nit: the `hpke_decrypt_test_schema.json` file has descriptions for this field + some others. Could you copy over those descriptions here too? They look equivalently applicable.

schemas/hpke_decrypt_test_schema.json
```diff
@@ -0,0 +1,371 @@
...
+          "enum": [
+            "HpkeTest"
+          ],
+          "description": "A legacy field, test files contain only one type per file"
+        },
+        "source": {
+          "$ref": "common.json#/definitions/Source"
+        },
+        "kem": {
+          "enum": [
```
cpu, line 64: This schema has a more exhaustive `kem` enum that includes algorithms that don't have vectors present in this branch. The `encap` equivalent only lists enum variants for KEMs that have vectors. 

We should probably be consistent in approach here and I lightly favor only listing KEMs that have vectors. WDYT?

doc/hpke.md
```diff
@@ -0,0 +1,159 @@
...
+The reject reason is KEM-specific:
+
+- **ML-KEM**: the encapsulation-key check (FIPS 203 Section 7.2) rejects a
+  coefficient at or past the modulus, or the key has the wrong length. An
+  all-zero ML-KEM key is `valid` — every coefficient is 0, below the modulus,
+  so the check passes and `SetupBaseS` must succeed; rejecting it is
+  over-strict.
+- **DHKEM(X25519)**: HPKE is stricter than plain X25519 here. A small-order
+  `pkRm` whose Diffie-Hellman output is all-zero MUST be rejected, because
+  RFC 9180 Section 5.1.1 requires `DH` to abort on an all-zero shared secret —
```
cpu, line 46: I think [RFC 9180 §7.1.4](https://www.rfc-editor.org/rfc/rfc9180.html#name-validation-of-inputs-and-ou) "Validation of Inputs and Outputs" is the right citation for this, no? 

[§5.1.1](https://www.rfc-editor.org/rfc/rfc9180.html#section-5.1.1) as cited here is "Encryption to a Public Key" and doesn't discuss the all-zero shared secret while 7.1.4 explicitly says:

> In particular, recipients MUST check whether the Diffie-Hellman shared secret is the all-zero value and abort if so.

If you agree with me then the same citation needs to be fixed in a few other spots in this branch.

testvectors_v1/hpke_mlkem1024_decrypt_test.json
```diff
@@ -0,0 +1,530 @@
...
+      "description": "enc and ct are a valid encapsulation and sealing for a different recipient key. Decapsulating under this skRm yields a different shared secret, so Open must fail."
+    }
+  },
+  "testGroups": [
+    {
+      "type": "HpkeTest",
+      "source": {
+        "name": "wycheproof",
+        "version": "0.1"
+      },
```
cpu, line 110: Can you make the source name more specific throughout? I think something like `github/sgmenda/hpke-wycheproof` would be better if you're open to it.

doc/hpke.md
```diff
@@ -0,0 +1,159 @@
...
+| `decryptions[].result` | `valid` if this `Open` must succeed, `invalid` if it must fail. |
+
+### Additional Test Vectors
+
+None yet. Every current case is generated by Wycheproof (`source.name =
+"wycheproof"`). The published specifications carry their own worked examples —
+RFC 9180 Appendix A for the classical DHKEM ciphersuites, and the
+[draft-ietf-hpke-pq][] test vectors for the ML-KEM KEMs — which would make a
+good independent cross-check. Importing them (with `source` pointing at the
+spec) is a planned addition; this section is a placeholder until then.
```
cpu, line 156: I'm not sure we need this placeholder text.

schemas/hpke_decrypt_test_schema.json
```diff
@@ -0,0 +1,371 @@
...
+              },
+              "flags": {
+                "type": "array",
+                "items": {
+                  "type": "string"
+                },
+                "description": "A list of flags"
+              }
+            },
+            "required": [
```
cpu: I think usually it's the object with the `tcId` that has the required `result`, but in this case it's per `decryptions` - is that right? Are there any cases where the decryptions aren't a consistent `result` across the board?

schemas/hpke_decrypt_test_schema.json
```diff
@@ -0,0 +1,371 @@
...
+              },
+              "pkRm": {
+                "type": "string",
+                "format": "HexBytes",
+                "description": "The serialized recipient public key (SerializePublicKey of pkR), derivable from skRm and provided for convenience"
+              },
+              "ikmE": {
+                "type": "string",
+                "format": "HexBytes",
+                "description": "Optional. The 32-byte seed that fixes the sender's ephemeral key. For ML-KEM it is the encapsulation randomness (ikmE); for DHKEM(X25519) it is the sender's ephemeral private key skEm, i.e. the seed a deterministic-encapsulation test hook consumes, not RFC 9180 DeriveKeyPair ikm. When present, deterministic encapsulation from this seed reproduces enc and ct exactly, so an implementation with such a hook can also check the sender direction. Absent for negative cases, where enc/ct are mutated and are not reproducible."
```
cpu, line 133: This feels a little bit awkward to me because it results in an `ikmE` with semantics that differ based on the algorithm. For the Go stdlib HPKE the ML-KEM cases were OK, but for the X25519 cases I couldn't just feed the value through our derandomization helper because it calls `DeriveKeyPair` and that's already been done, instead I have to switch on the algorithm and directly use `NewPrivateKey` for the bytes for X25519.

ISTM that putting the RFC 9180 ikmE value here across the board is more general and BoringSSL-family consumers can do the derivation to get skEm themselves (and/or we could have an optional field alongside the ikmE that has the expected skEM for them).

WDYT?

testvectors_v1/hpke_x25519_decrypt_test.json
```diff
@@ -0,0 +1,475 @@
...
+    "ModifiedTag": {
+      "bugType": "AUTH_BYPASS",
+      "description": "A bit of the AEAD authentication tag was flipped."
+    },
+    "Pseudorandom": {
+      "bugType": "BASIC",
+      "description": "A valid HPKE base-mode ciphertext with deterministic encapsulation from ikmE (the sender's ephemeral private key)."
+    },
+    "SmallOrderEnc": {
+      "bugType": "CAN_OF_WORMS",
```
cpu, line 74: The counterpart ZeroSharedSecret note on the encap tests use DEFINED for this. Should we be consistent here?

cpu (commented):
Thanks! I think these are almost ready to go, I just had a few items of feedback to consider first. 

I was able to write a downstream test using this branch for the Go HPKE impl of ML-KEM-768, ML-KEM-1024, & X25519 (we don't have ML-KEM-512) and everything produced the expected results. 👍

PR #274, opened 2026-08-24
mlkem: add re-encryption comparison vectors to the seed-key files

007bsd:
Adds message-preserving ML-KEM re-encryption test vectors to the seed-key files
(mlkem_{512,768,1024}_test.json): each fails to decapsulate on a correct
implementation but returns the real shared secret on one that skips part of the
ciphertext comparison.

#254 added this kind of vector only to the semi-expanded decaps files; the
seed-key files have none (their existing reject vectors change the message, so
they cannot isolate a single-region skip). This adds one MLKEMTest group per
parameter set (7/8/9 tests), one single-byte modification per structural region:
byte 0, each c1 block boundary, and c2 start / interior / N-32 / last.

Motivated by the incomplete-comparison class behind CVE-2026-10097 and
CVE-2026-6330 (public, fixed); submitted post-fix per CONTRIBUTING. Generator and
cross-checks (OpenSSL, wolfSSL, Go crypto/mlkem):
https://github.com/007bsd/mlkem-implicit-rejection


cpu (approved):
Thanks! This looks great.

Did you have any feedback about the process & `vectorgen`? Was it relatively painless? 🤞

007bsd:
> Thanks! This looks great.
> 
> Did you have any feedback about the process & `vectorgen`? Was it relatively painless? 🤞

Thanks! Along with the CVEs above, these also cover the RustSec advisory I filed ([#3151](https://github.com/rustsec/advisory-db/pull/3151)) -- same implicit-rejection class. 

The process was smooth. My only real question was whether these belonged with the decap vectors or in their own group. We went with a separate group, and vectorgen was easy to work with from there.

cpu:
I'm going to merge this but if anyone shows up with feedback after the fact we can iterate further.

Thanks for the contribution!

